PR #12, opened 2015-11-14
WIP More Appveyor tests

bsipocz:
I have to open it to see whether the the changes work. 

Note: Here is an online powershell terminal, so some random stuff can be tried out before committing: http://www.tutorialspoint.com/powershell_terminal_online.php

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,18 @@ environment:
       - PYTHON_VERSION: "3.5"
         NUMPY_VERSION: "1.10"
 
+      - PYTHON_VERSION: "3.4"
+        NUMPY_VERSION: "1.9"
+        ASTROPY_VERSION: "development"
+
+      - PYTHON_VERSION: "3.4"
+        NUMPY_VERSION: "1.9"
+        ASTROPY_VERSION: "stable"
+
+      - PYTHON_VERSION: "3.4"
+        NUMPY_VERSION: "1.9"
+        ASTROPY_VERSION: "1.0"
+
 
 platform:
     -x64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,23 +15,30 @@ environment:
 
       - PYTHON_VERSION: "2.6"
         NUMPY_VERSION: "1.9"
+        TEST_CMD: "import numpy; assert numpy.__version__.startswith(\"1.9\")"
+
       - PYTHON_VERSION: "3.4"
         NUMPY_VERSION: "1.9"
+        TEST_CMD: "import numpy; assert numpy.__version__.startswith(\"1.9\")"
+
       - PYTHON_VERSION: "3.5"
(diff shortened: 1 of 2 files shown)


astrofrog:
@bsipocz - thanks for working on that! By the way, if you want to avoid having to cancel the travis tests, you can also just remove the travis file temporarily.

bsipocz:
@astrofrog - thanks, that's a good idea to remote the travis file. Also I won't be able to finish this today, but wanted so see whether it works at all or not (looks like the latter).

bsipocz:
@astrofrog - It's weird travis still starts up for a new commit, but obviously fails as there's no ``.travis.yml`` file

appveyor.yml
```diff
@@ -39,9 +62,10 @@ install:
...
 
-# Not a .NET project, we build SunPy in the install step instead
+# Not a .NET project, we build ci-helpers in the install step instead
 build: false
 
 test_script:
-  - "%CMD_IN_ENV% python --version"
-
+  - "echo %TEST_CMD%"
+  - "%CMD_IN_ENV% python -c %TEST_CMD%"
```
astrofrog: I wonder if the issue is that there are quotes missing arond ``%TEST_CMD``? I.e. at the moment it's doing something like

```
python -c import numpy ...
```

instead of

```
python -c 'import numpy ...'
```

?

appveyor.yml
```diff
@@ -13,14 +13,37 @@ environment:
...
+
+      - PYTHON_VERSION: "3.5"
+        NUMPY_VERSION: "1.10"
+        ASTROPY_VERSION: "1.0"
+        TEST_CMD: 'import numpy; import astropy; assert numpy.all([numpy.__version__.startswith(\"1.10\"); astropy.__version__.startswith(\"1.0\")])'
+
+      - PYTHON_VERSION: "2.7"
+        NUMPY_VERSION: "1.10"
+        ASTROPY_VERSION: "1.0"
+        TEST_CMD: 'import numpy; import astropy; assert numpy.all([numpy.__version__.startswith(\"1.10\"), astropy.__version__.startswith(\"1.0\")])'
```
astrofrog: I wonder if it might help debugging to split up the asserts into ``assert numpy.startswith(...); assert astropy.startswith(...)``? Maybe also then change ``\"`` back to ``"``?

astrofrog: Ignore me, I just saw you were already trying different possibilities above :)

astrofrog:
@bsipocz - I had an idea for how we can do the testing more easily and avoid these kind of issues, I'll be on gitter today, so just ping me and we can discuss it.

bsipocz:
I've cleaned this up from the debugging tries as some of the commits were genuine and needed changes we may want to cherry-pick elsewhere.

bsipocz:
Closing as #14 picked up the commits from here.

PR #20, opened 2015-11-19
Make pip output quiet

astrofrog:
This should get rid of progress bars (which clutter CI logs otherwise). Need to check before merging.


astrofrog:
I'm not sure, it's a bit too quiet now - would be nice to just get rid of progress bar.

PR #27, opened 2015-11-29
Turn on bash debugging and print some extra info

### LINKED ISSUE #25: Show commands executed in setup scripts
Is it possible to change the setup so that the executed commands appear in the travis-ci log?

Currently only
```
$ source ci-helpers/travis/setup_conda_$TRAVIS_OS_NAME.sh
```
is displayed and then the output from the commands in that script, but not the commands themselves.
Example: https://travis-ci.org/astropy/ci-helpers/jobs/92443363#L268

This makes the logs very hard to read and makes it  difficult to debug issues if one of those commands fails or doesn't do what one expects. It's cost me a few hours over the past year and it's always frustrating when issues happen at that stage.

On the other hand, the travis-ci part of the build log is usually very easy to read because the commands are printed and even highlighted / grouped in the build log.

I'm not familiar much with shell scripting, but there should be a flag / option to run these scripts so that commands are echoed before they are executed, no?

bsipocz:
This is to address the discussion in #25. Closes #25.

ping @cdeil @astrofrog

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ Following this, various dependencies are installed depending on the following en
 * ``$CONDA_CHANNELS``: this should be a space-separated string of conda
   channel names, and defaults to ``astropy-ci-extras``.
 
+* ``$DEBUG``: if True this turn on the shell debug mode in the install scripts
+
 The idea behind the ``MAIN_CMD`` and ``SETUP_CMD`` environment variables is
 that the ``script`` section of the ``.travis.yml`` file can be set to:
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,8 @@ Following this, various dependencies are installed depending on the following en
 * ``$CONDA_CHANNELS``: this should be a space-separated string of conda
   channel names, and defaults to ``astropy-ci-extras``.
 
-* ``$DEBUG``: if True this turn on the shell debug mode in the install scripts
+* ``$DEBUG``: if `True` this turns on the shell debug mode in the install
+  scripts
 
 The idea behind the ``MAIN_CMD`` and ``SETUP_CMD`` environment variables is
 that the ``script`` section of the ``.travis.yml`` file can be set to:
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,12 @@ matrix:
                CONDA_CHANNELS='astropy-ci-extras astrofrog'
                CONDA_DEPENDENCIES='PyQt5'
 
+        - python: 3.5
+          env: SETUP_CMD='test' DEBUG=True
+
+        - python: 3.5
+          env: SETUP_CMD='test' DEBUG=False
+
 
 before_install:
     - ln -s . ci-helpers
--- a/travis/setup_conda_linux.sh
+++ b/travis/setup_conda_linux.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-echo "==================== Starting executing ci-helpers scripts ====================="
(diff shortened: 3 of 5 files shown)


cdeil:
Is seems that somehow sub-commands or sub-shell info are echoed as well.
And there's all kinds of extra lines printed after the shell script as well, i.e. the setting isn't turned off after exiting the shell script.
Example: https://travis-ci.org/astropy/ci-helpers/jobs/93804456

I think the behaviour from `.travis.yml` to exactly show the commands executed directly and then their output is best. Any idea how to get the same behaviour for commands from a shell script instead of `.travis.yml`? Maybe it's not possible?

bsipocz:
Yes, they are echoed as 'set -x' is debugging mode.

I don't know any other way to do this, and travis must have a separate
echoing function that echoes every line of the yml before it's passed on to
bash.

astrofrog:
I think if you use ``#!/bin/bash -x`` instead of ``set -x`` it shows only the commands in that shell script and not all the extra stuff.

bsipocz:
Yes, it's already there but had no affect. I'll try to run in differently, maybe it will work that way.

bsipocz:
It only got [worse](https://travis-ci.org/astropy/ci-helpers/builds/93919764) by changing the source to direct running of the script as it messed up something and now the packages cannot be found even that they are installed.

Anyway, doing that way using ``#!/bin/bash -x`` only, gives the same output. The subscript (activate test) is debugged as it is being "sourced" into the main script. I think the only workaround is to switch off debugging right before sourcing it, and turn it back on after the subscript has finished.

bsipocz:
Today I run into part of this PR, where the ``set -e`` would have been helpful. 
Thus these two questions to @astrofrog and @cdeil:

- Do you think we should hard wire it into ``ci-helpers`` to fail the build when a non-zero exit status comes out of the build shell script? (I think it's good, as it means something gone wrong during the installation, and saves a headache to start the dead-end investigating at the failing tests levels)

- What do you think about the whole PR?  Should we merge it, or close it?

astrofrog:
> Do you think we should hard wire it into ci-helpers to fail the build when a non-zero exit status comes out of the build shell script? (I think it's good, as it means something gone wrong during the installation, and saves a headache to start the dead-end investigating at the failing tests levels)

Yes

> What do you think about the whole PR? Should we merge it, or close it?

I'm not sure - I find the build output a bit noisy with the full output, but I don't have a strong opinion.

astrofrog:
We could always have a global ``DEBUG`` variable that some packages can choose to opt in to to get more detailed output.

astrofrog:
@bsipocz - this looks good! The scripts are missing a newline at the end, but otherwise this looks like a good solution to me. However, can you also add a mention of this in the README?

bsipocz:
Sure, will do this tomorrow.

bsipocz:
@astrofrog - A line is added in the readme. Also I kept the two description lines in for all the builds.

README.md
```diff
@@ -81,6 +81,8 @@ Following this, various dependencies are installed depending on the following en
 * ``$CONDA_CHANNELS``: this should be a space-separated string of conda
   channel names, and defaults to ``astropy-ci-extras``.
 
+* ``$DEBUG``: if True this turn on the shell debug mode in the install scripts
```
astrofrog: put backticks around ``True``, and turn -> turns

astrofrog:
:+1: apart from one minor comment above

PR #32, opened 2015-12-29
Update install-miniconda.ps1

embray:
Change confusing variable name--I expected `python_version` to be something like "2.7" or "3.4", whereas this is the version of the miniconda distributable.  I'm tempted also to change `python_home` to `miniconda_home`, but that is filled below from an environment variable called just "PYTHON" which I'm tempted to also rename to "CONDA" or something.

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -43,8 +43,8 @@ function DownloadMiniconda ($version, $platform_suffix) {
    return $filepath
 }
 
-function InstallMiniconda ($python_version, $architecture, $python_home) {
-    Write-Host "Installing miniconda" $python_version "for" $architecture "bit architecture to" $python_home
+function InstallMiniconda ($miniconda_version, $architecture, $python_home) {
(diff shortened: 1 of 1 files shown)


bsipocz:
@embray - Thanks for this. Please go ahead with renaming the other variables, too.

bsipocz:
Also since #33 got merged, this should be rebased (we require that all tests pass before being able to merge, thus PRs must be rebased after each merge).

embray:
I'm hesitant to change the other variable because of lines like this:
```
InstallMiniconda $env:MINICONDA_VERSION $env:PLATFORM $env:PYTHON
```
What irks me is that here `$env:PYTHON` points to a miniconda installation, which is more than just a Python installation.  But that `$env:PYTHON` is set by each projects' appveyor.yml, which means if I change the name of this variable astropy and all affiliated packages will have to be updated yet again.  I think it's not terribly important though.  It's still the path to the Python distribution, so it makes some sense (compared to "python_version").

I will rebase though.

bsipocz:
Thanks again @embray.

PR #34, opened 2015-12-30
Set up Windows SDK when calling pip install

embray:
pip installs should be preceded by the `%CMD_IN_ENV%` invokation to set up the Windows SDK environment when appropriate, since pip install will often (not always, depending on wheel availability) result in a build from source, in which case we presumably want to build for the same platform as the package that we're testing.

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -130,9 +130,9 @@ conda install -n test -q pytest $NUMPY_OPTION $ASTROPY_OPTION $CONDA_DEPENDENCIE
 
 # Check whether the developer version of Numpy is required and if yes install it
 if ($env:NUMPY_VERSION -match "dev") {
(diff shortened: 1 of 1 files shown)


bsipocz:
@embray - This needs a rebase and then we can merge it.

 (btw is it possible for a maintainer to rebase a PR somehow and push it back to the same PR?)

embray:
@bsipocz I just clicked the (new?) "update branch" button to merge it into the latest master, so it shouldn't need to be rebased now.

bsipocz:
Yes, you did it while I was writing the comment :)

embray:
Cool--FWIW I was just now able to confirm locally that this resolved the issue.  Invoking pip through the windows_sdk.cmd script successfully set up the environment for using the Windows SDK (of which I also needed to install the correct version), and built astropy dev with all the right flags.

PR #38, opened 2016-01-09
Travis add mpl backend setup

### LINKED ISSUE #26: Add mpl setup as an opt-in
Comment from @cdeil:
"Many packages need this. Move this to ci-helpers and make it opt-in, i.e. execute these commands if the user sets something like SETUP_DISPLAY_FOR_MATPLOTLIB?"

This refers to the setup needed for mpl:
```
# Make sure that interactive matplotlib backends work
export DISPLAY=:99.0;
sh -e /etc/init.d/xvfb start;
```

What do you think @astrofrog? Would this work for both the linux an osx builds? For the latter I saw packages prefer to use pyqt rather than pyside, but do we want hardwire such things?

bsipocz:
This is to address and close #26.

@astrofrog - do we need to test this somehow?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,6 @@ Following this, various dependencies are installed depending on the following en
 * ``$DEBUG``: if `True` this turns on the shell debug mode in the install
   scripts, and provides information on the current conda install.
 
-* ``$SETUP_MPL``: if True this makes sure the interactive matplotlib
-  backends work
-
 * ``$PACKAGENAME_VERSION``: ``PACKAGENAME`` is the name of the package to
   specify the version for (e.g. ``MATPLOTLIB_VERSION``). Due to shell
   limitations, all hyphens in the conda package name should be changed to
--- a/README.md
+++ b/README.md
@@ -92,6 +92,9 @@ Following this, various dependencies are installed depending on the following en
 * ``$DEBUG``: if `True` this turns on the shell debug mode in the install
   scripts, and provides information on the current conda install.
 
+* ``$SETUP_MPL``: if True this makes sure the interactive matplotlib
+  backends work
+
 * ``$PACKAGENAME_VERSION``: ``PACKAGENAME`` is the name of the package to
   specify the version for (e.g. ``MATPLOTLIB_VERSION``). Due to shell
   limitations, all hyphens in the conda package name should be changed to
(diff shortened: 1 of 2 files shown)


astrofrog:
Of course, we could also simply always run those commands - I'm not sure if there would be any downsides?

astrofrog:
@bsipocz - I think we should actually just always run these commands, BUT we need to check that the OS is linux.

bsipocz:
@astrofrog - Rebased and removed the env variable so it's always set. 
As it's only included in ``setup_conda_linux.sh``, the OS should be linux.

bsipocz:
Rebased again. And I've also sent a feature request to support@github whether they would consider adding a "Rebase branch" button next to the current "Update branch". Latter merges master back to the given feature branch which we don't like that much.

bsipocz:
@astrofrog - Do you know how to disable the coveralls here, so it can't screw up the statuses? It doesn't make much sense anyway as the only python here is the test itself, and we only run it to make sure we installed all its dependencies.

astrofrog:
@bsipocz - I've disabled the coverage status (it's an option on coveralls.io)

bsipocz:
Thanks.

bsipocz:
Rebased on #45 so a new build is triggered (and that one should go in asap as it already causes trouble that the package names are not always lowercase (eg Cython))

astrofrog:
@bsipocz - it looks like this will need rebasing

bsipocz:
I'll rebase once we merge #51 (as a rebase is needed after every merge).

astrofrog:
@bsipocz - sounds good!

bsipocz:
Rebased

bsipocz:
Merged this to avoid YAR (yet another rebase) ;)

astrofrog:
Thanks! I just discovered that packages will need to remove these lines otherwise the Travis build fails. I'll email the mailing list for affiliated package maintainers.

bsipocz:
Ouch. Should we use with an env variable then?

bsipocz:
Also some packages are checking for qt, while others seems to be ok with pyside. Should be put that here, too?

astrofrog:
@bsipocz - ok, maybe it makes sense to add an environment variable ``XVFB`` for people to opt in. Could you do that? I wouldn't touch the ``QT_API`` variable which some packages set though, that's more specialized (is that what you meant?)

bsipocz:
Yes, I'm doing it now. And yes, that's what I meant, that we cannot do it being generic enough anyway to make all the affiliated packages happy.

PR #40, opened 2016-01-12
Removing conda info command from the build script

bsipocz:
as it's really only for debugging, and the warnings it generates messes up the sphinx builds warning catching system.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -139,7 +139,7 @@ fi
 
 if [[ $DEBUG == True ]]; then
     # include debug information about the current conda install
-    $CONDA_INSTALL _license
+    conda install -n root _license
     conda info -a
 fi
 
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -22,7 +22,6 @@ do
 done
 
 conda update -q conda
-conda info -a
 
 # Use utf8 encoding. Should be default, but this is insurance against
 # future changes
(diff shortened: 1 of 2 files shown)


bsipocz:
Now the ``conda info -a`` is put back when ``DEBUG=True``. Also installs the _license in this case as it's missing state was the cause of the WARNING.

bsipocz:
@astrofrog - Is this OK with you?

travis/setup_dependencies_common.sh
```diff
@@ -138,4 +137,10 @@ if [[ $ASTROPY_VERSION == dev* ]]; then
     $PIP_INSTALL git+http://github.com/astropy/astropy.git#egg=astropy --upgrade
 fi
 
+if [[ $DEBUG == True ]]; then
```
astrofrog: Just curious - do you need ``== True``?

astrofrog:
:+1:

travis/setup_dependencies_common.sh
bsipocz: I think yes as it only tests for extistence and not the bool value, so it's always True for any value DEBUG is set.

```
bsipocz@tegla2:~/munka/devel/ci-helpers$ a=False
bsipocz@tegla2:~/munka/devel/ci-helpers$ if [[ $a ]]; then echo 1;fi
1
bsipocz@tegla2:~/munka/devel/ci-helpers$ a=qqriq
bsipocz@tegla2:~/munka/devel/ci-helpers$ if [[ $a ]]; then echo 1;fi
1
```

astrofrog: Ok, sounds good :)

PR #41, opened 2016-01-13
Fix matplotlib version to 1.5.0 for travis-ci sphinx builds

larrybradley:
`travis-ci` sphinx builds will fail with `matplotlib 1.5.1` because of a warning about `fc-list`.  This PR can be reverted once https://github.com/matplotlib/matplotlib/issues/5836 is fixed.


larrybradley:
I changed this from `pip` to `conda` which should also work.  However, `matplotlib` will need to be removed as a conda dependency for the Sphinx builds in the package `.travis.yml` to prevent `matplotlib` from being updated to `1.5.1`.

bsipocz:
Merging this, so packages that rely on matplotlib wont get the failure.

PR #43, opened 2016-01-13
Pinning dependency versions

### LINKED ISSUE #42: Pinning packages version number to aviod upgrade
We had the issue that we coundn't fix the mpl version number to ``1.5.0`` in the photutils docs builds (https://github.com/astropy/photutils/pull/318) as it always got upgraded.

The upgrade is due to https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L106 where mpl isn't inheriting the fixed version number.

We can have temporary workarounds like #41, or introduce version number enviroment variables (like we have for astropy and numpy alread) for all the packages that get installed as an instrastructural dependency (like sphinx and matplotlib).

However it may be the best to pin the version numbers of all the packages that are listed with a required version number in ``CONDA_DEPENDENCIES``. I have a half baked hack for this, and will open a PR once I manage to get it work.

cc: @larrybradley @astrofrog

bsipocz:
This is to fix #42 (in favour of #41, so closes #41).

@astrofrog - What do you think? Should we introduce an env variable to opt into this, or can this be the behaviour always?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,11 @@ matrix:
                TEST_CMD="python -c 'import matplotlib; assert matplotlib.__version__==\"1.5.0\"'"
 
         - os: linux
+          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0'
+               MATPLOTLIB_VERSION=1.5.1
+               TEST_CMD="python -c 'import matplotlib; assert matplotlib.__version__==\"1.5.1\"'"
+
+        - os: linux
           env: SETUP_CMD='build_docs' PYTHON_VERSION=2.7
 
         - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,7 @@ matrix:
 
         - os: linux
           env: SETUP_CMD='test --coverage' PYTHON_VERSION=2.7
+               TEST_CMD='coverage run test_env.py'
                CONDA_CHANNELS='astropy-ci-extras astrofrog'
                CONDA_DEPENDENCIES='pyqt5' DEBUG=True
                NUMPY_VERSION=1.10 ASTROPY_VERSION=lts
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
 
         - os: linux
           env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0'
-               TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=='1.5.0'"
+               TEST_CMD="python -c 'import matplotlib; assert matplotlib.__version__==\"1.5.0\"'"
 
         - os: linux
           env: SETUP_CMD='build_docs' PYTHON_VERSION=2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,8 @@ matrix:
                CONDA_DEPENDENCIES='scipy'
 
         - os: linux
-          env: SETUP_CMD='build_sphinx'
+          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0'
+               TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=='1.5.0'"
 
         - os: linux
           env: SETUP_CMD='build_docs' PYTHON_VERSION=2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ matrix:
         - os: linux
           env: SETUP_CMD='test' PYTHON_VERSION=2.7
                CONDA_CHANNELS='astropy-ci-extras astrofrog'
-               CONDA_DEPENDENCIES='PyQt5'
+               CONDA_DEPENDENCIES='pyqt5' DEBUG=True
 
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,12 +45,12 @@ matrix:
         - os: linux
           env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0'
                DEBUG=True
-               TEST_CMD="python -c 'import matplotlib; assert matplotlib.__version__==\"1.5.0\"'"
+               TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=="\""1.5.0"\"'
 
         - os: linux
-          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0'
+          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0 sphinx'
                MATPLOTLIB_VERSION=1.5.1 DEBUG=True
-               TEST_CMD="python -c 'import matplotlib; assert matplotlib.__version__==\"1.5.1\"'"
+               TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=="\""1.5.1"\"'
 
         - os: linux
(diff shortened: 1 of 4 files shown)


astrofrog:
I'll take a look at this tomorrow - thanks for working on it! :)

astrofrog:
@bsipocz - could you add a comment with a link to documentation about the pinned feature of conda?

Another cool thing we could do would be to auto-detect environment variables that end in ``_VERSION`` and use that to pin versions of packages. Specifically, imagine you have

```
CONDA_DEPENDENCIES='astropy matplotlib pandas yt'
```

it would be cool to be able to set e.g.

```
MATPLOTLIB_VERSION=1.5
```

for one build and 

```
MATPLOTLIB_VERSION=1.4
```

for another without having to change ``CONDA_DEPENDENCIES``. Maybe we can loop over the packages in the conda and pip list as you are doing at the moment, and if the version is not specified, check if an environment variable named package.upper() + '_VERSION' exists, and if so, use that.

What do you think?

bsipocz:
@astrofrog - this idea with the version env variable checking sounds good, and can save a lot of headache.

astrofrog:
@bsipocz - thanks for working on this! I wonder whether it would make sense to just raise an error if the version number is specified both in the list and in an environment variable?

astrofrog:
@bsipocz - just a thought - if the bash script becomes too complex, there's also the option of just calling a Python script to do the bit with the version numbers (since the Python script could access the env variables too). It would make it easier to set up a dictionary of version numbers.

bsipocz:
@astrofrog - The bash script is still not that bad, I just made silly mistakes (as it's difficult to test this locally). But now it should really work (I hope :) ).

bsipocz:
@astrofrog - This is ready now. The one test failure seams unrelated, but it's absent from the master (even when I rerun the tests on master now), so I'm not sure exactly why it's happening. An output with the debug option is here:
https://travis-ci.org/bsipocz/ci-helpers/jobs/102664653

The channel  ``astrofrog`` is added to the config here: https://travis-ci.org/bsipocz/ci-helpers/jobs/102664653#L317, so ``conda install -q python=2.7 PyQt5`` should work in https://travis-ci.org/bsipocz/ci-helpers/jobs/102664653#L469

astrofrog:
@bsipocz - it looks like the PR won't be mergeable because the last commit has ``[ci skip]`` - can you edit the last commit message and remove ``[ci skip]``?

astrofrog:
@bsipocz - could the error be related to this message?

```
Note that you have pinned specs in /home/travis/miniconda/envs/test/conda-meta/pinned:
    ['PyQt5']
```

Since we aren't pinning the version of ``PyQt5``, it shouldn't appear in ``pinned``, right?

bsipocz:
It shouldn't be a problem to list it in the pin file, since no version is
given. Besides many other tests have passed with this approach. I'll try
one more hing and if it still doesn't work then I'll clean up then pin file
to contain only theversioned packages.

astrofrog:
Maybe the pinned version file doesn't work with packages that are not from the default channel?

.travis.yml
```diff
@@ -92,12 +99,6 @@ matrix:
                CONDA_CHANNELS='astropy-ci-extras astrofrog'
                CONDA_DEPENDENCIES='PyQt5'
```
astrofrog: @bsipocz - the issue is that this should actually be ``pyqt5`` (all lowercase). Not sure how it ended up uppercase and how it worked before.

bsipocz:
@astrofrog - Now it passes everything. Could you give it a final review?

README.md
```diff
@@ -72,8 +72,10 @@ Following this, various dependencies are installed depending on the following en
   version is installed (more info about LTS can be found
   [here](https://github.com/astropy/astropy-APEs/blob/master/APE2.rst#version-numbering).
 
-* ``$CONDA_DEPENDENCIES``: this should be a space-separated string of package
-  names that will be installed with conda.
+* ``$CONDA_DEPENDENCIES``: this should be a space-separated string of
+  package names that will be installed with conda. Version numbers of these
+  dependencies can be overriden/specified with the ``$${PACKAGENAME}_VERSION``
+  environmental variables.
```
astrofrog: environment

README.md
```diff
@@ -72,8 +72,10 @@ Following this, various dependencies are installed depending on the following en
   version is installed (more info about LTS can be found
   [here](https://github.com/astropy/astropy-APEs/blob/master/APE2.rst#version-numbering).
 
-* ``$CONDA_DEPENDENCIES``: this should be a space-separated string of package
-  names that will be installed with conda.
+* ``$CONDA_DEPENDENCIES``: this should be a space-separated string of
+  package names that will be installed with conda. Version numbers of these
+  dependencies can be overriden/specified with the ``$${PACKAGENAME}_VERSION``
```
astrofrog: I think just saying '``$PACKAGENAME_VERSION``, where ``PACKAGENAME`` is the name of the package to specify the version for (e.g. ``MATPLOTLIB_VERSION``)' would be fine (i.e. no need for double ``$``)

travis/setup_dependencies_common.sh
```diff
@@ -51,6 +51,41 @@ if [[ $MAIN_CMD == pep8* ]]; then
...
+
+    # Let env variable version number override this pinned version
+    for package in $(gawk '{print $1}' $pin_file); do
+        version=$(eval echo -e \$$(echo $package | \
+            gawk '{print toupper($0)"_VERSION"}'))
+        if [[ ! -z $version ]]; then
+            gawk -v package=$package -v version=$version \
+                '{if ($1 == package) print package" " version; else print $0}' \
+                $pin_file > /tmp/pin_file_temp
+            mv /tmp/pin_file_temp $pin_file
```
astrofrog: Doesn't this overwrite the file each time?

bsipocz: Yes it does, but this seemed to be the simplest way to do this (rather than doing some sed ninja stuff (and even that would overwrite the file as I can't see an easy way out of the outer loop). Besides normally there aren't too many dependencies.

astrofrog: I guess I mean won't this override previous pinned dependencies? That is, it runs for one package, and then for the next package it overwrites it. Or am I mis-reading?

bsipocz: Nope, the conditional is inside gawk, so it only changes the pin file in the package's line and leaves the rest alone.

astrofrog: Sorry, am being slow today... But the ``mv`` command is in the if statement but not in gawk, right? So doesn't ``mv`` get called multiple times?

bsipocz: It's just bash stupidity, the input and output of the gawk should be different, thus I put everything into a temp file, and after the gawk overwrite the original input with it, then the new ``package`` in the loop starts it from the beginning. So you're right, it's super non-optimal for io but maybe the bash hack is more readable than otherwise it would be.
However this conditional only runs when there is any ``$PACKAGE_VERSION`` variable given, as we said it should override the version numbers listed in ``CONDA_DEPENDENCIES``.

bsipocz: Maybe it's easier to see in the travis log here https://travis-ci.org/astropy/ci-helpers/jobs/103374706#L455 what's going on. There is a ``MATPLOTLIB_VERSION``, but no ``SPHINX_VERSION``, so the pinned mpl 1.5.0 gets changed to 1.5.1.

astrofrog: Ok, I see now :) My bash-fu is not optimal...

astrofrog:
This is great :) Merging!

travis/setup_dependencies_common.sh
bsipocz: :) mine is rusty, too but at some point in the past that was my ultimate (and only) forte.

PR #45, opened 2016-01-19
Making sure all conda pendenency package is written lowercase 

bsipocz:
Conda supposed to be lowercase only anyway.


bsipocz:
Removed a few duplicate like builds from the appveyor runs as well as it seems that a backlog is already starting there as all the astropy and affiliated running in the same queue.

bsipocz:
@astrofrog  - merged as this should fix the errored builds of astropy. Could you please restart them tomorrow?

PR #46, opened 2016-01-20
Making sure all conda pendenency package is written lowercase

bsipocz:
in the pinfile, too. This should have been part of #45.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ matrix:
 
         - os: linux
           env: SETUP_CMD='test' ASTROPY_VERSION=stable
+               CONDA_DEPENDENCIES='Cython'
 
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=1.9 CONDA_DEPENDENCIES='requests'
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,10 +64,10 @@ matrix:
 
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=1.10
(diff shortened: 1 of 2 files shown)


bsipocz:
@hamogu - could you restart your failing travis build? This PR should have fixed it.

hamogu:
It works! Thanks!

embray:
This seems to have caused various builds for Astropy to break :/

bsipocz:
@embray - Yes, pinning the version numbers in #43 caused more bugs in corner cases than it solved. The fix is coming in #48, just need to wait for appveyor and travis.

hamogu:
Do we need a ci-helpers-test package, that is cloned when astropy-test-package (to be written) and astropy-affiliated-test-package (to be written) are build?
And then ci-helpers would be a clone of the last commit of ci-helpers-test that passed with a astropy-test-pacakge and astropy-affiliated-test-package ...

Arg, my head hurts from thinking about so many levels of nested testing the tests...

bsipocz:
Or we just need to be even more careful with the tests here, to make sure they cover all the extra cases (e.g conda does some tricks with package names and version numbers when it's in the command line, but it requires the pin file to be more pedantic (e.g conda packages are always lowercase, still it passes with mixed cases, or extends invalid version numbers, e.g it silently treats numpy=1.10 as numpy=1.10*, etc.). Some of these caught in the original PR, but sadly some slipped through. 
I even though that gawk may be a problem and should use awk instead, but didn't make it a test case.

PR #51, opened 2016-01-21
Have a way to provide options to conda

astrofrog:
In Glue, I was passing flags to conda using e.g.

```
CONDA_DEPENDENCIES='numpy scipy --no-deps'
```

this is now broken BUT I think that's fine, since this isn't really a proper way to do it. So I would suggest we add a new variable ``$CONDA_FLAGS`` to deal with passing any flags such as this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,10 +81,10 @@ Following this, various dependencies are installed depending on the following en
   names that will be installed with pip.
 
 * ``CONDA_DEPENDENCIES_FLAGS``: additional flags to pass to conda when
-  instlaling ``CONDA_DEPENDENCIES``
+  installing ``CONDA_DEPENDENCIES``
 
 * ``PIP_DEPENDENCIES_FLAGS``: additional flags to pass to pip when
-  instlaling ``PIP_DEPENDENCIES``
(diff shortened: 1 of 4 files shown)


astrofrog:
I've attached some code :)

astrofrog:
@bsipocz - note that I called the variables ``CONDA_DEPENDENCIES_FLAGS`` and ``PIP_DEPENDENCIES_FLAGS`` to make it clear the flags are just used when installing ``CONDA_DEPENDENCIES``, not for all conda commands.

README.md
```diff
@@ -80,6 +80,12 @@ Following this, various dependencies are installed depending on the following en
 * ``$PIP_DEPENDENCIES``: this should be a space-separated string of package
   names that will be installed with pip.
 
+* ``CONDA_DEPENDENCIES_FLAGS``: additional flags to pass to conda when
+  instlaling ``CONDA_DEPENDENCIES``
```
bsipocz: typo here and two lines below

astrofrog: Oops, fixed

bsipocz:
Looks good to me.

Issue #52, opened 2016-01-22
Specifying CONDA_DEPENDENCIES vs environment variables


astrofrog:
The new system of specifying package versions via environment variables works nicely. However, there is a small difference in behavior between specifying them via environment variables and via the ``CONDA_DEPENDENCIES``, i.e.

```
MATPLOTLIB_VERSION=1.3
```

vs

```
CONDA_DEPENDENCIES='matplotlib=1.3'
```

The first case results in the pinned file containing

```
matplotlib 1.3*
```

and the second:

```
matplotlib ==1.3
```

I think that the second case should also be changed to mean ``1.3*``, because that's normally how it works if we do ``conda install matplotlib=1.3`` (1.3.1 could be installed)

bsipocz:
Done in #53.

Issue #50, opened 2016-01-20
Add option to remove ``-q`` flag for conda


astrofrog:
In some cases, people might want to not have conda builds be quiet, so maybe we can add an env variable (or simply re-use ``$DEBUG``) to indicate that we want verbose output from conda.

bsipocz:
Either way is fine by me, although I'm a bit inclined to reuse DEBUG and put this into that, too.

bsipocz:
Done in #53

PR #53, opened 2016-01-22
Various enhanchements for the travis builds

bsipocz:
- we should handle alphanumeric package names when pinning the versions
- fixing #52 by adding ``*`` to the end of all lines with ``==`` in the pin file
- addressing #50 with building the ``-q`` conda flag into the $DEBUG variable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,8 @@ matrix:
           env: SETUP_CMD='test' ASTROPY_VERSION=stable
 
         - os: linux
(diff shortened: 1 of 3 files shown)


astrofrog:
@bsipocz - thanks! can you add a test that includes ``CONDA_DEPENDENCIES='matplotlib=1.5'`` and make sure that the version installed is at least 1.5.1 or above?

bsipocz:
@astrofrog - Added.

PR #54, opened 2016-01-22
Revert "Travis add mpl backend setup"

bsipocz:
Reverts astropy/ci-helpers#38

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,9 @@ Following this, various dependencies are installed depending on the following en
 * ``$DEBUG``: if `True` this turns on the shell debug mode in the install
   scripts, and provides information on the current conda install.
 
+* ``$SETUP_MPL``: if True this makes sure the interactive matplotlib
+  backends work
+
 * ``$PACKAGENAME_VERSION``: ``PACKAGENAME`` is the name of the package to
   specify the version for (e.g. ``MATPLOTLIB_VERSION``). Due to shell
   limitations, all hyphens in the conda package name should be changed to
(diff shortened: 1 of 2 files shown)


astrofrog:
Instead of reverting, what about having an ``$XVFB`` environment variable that can be set to True? I think that would be already useful, since it's annoying to have to remember those two lines, and if they ever need to change in future, will be easier to do from one place?

astrofrog:
(what I meant was that PySide vs Qt was too specialized, but still think that XVFB would be worthwhile)

bsipocz:
@astrofrog - That's what I try to do, to keep the first commit from the original, but revert the second one that made it an always on feature. This is the first time I tried the revert function, so maybe it's not ideal way to do it.

astrofrog:
@bsipocz - ok, no problem :) I would call the variable ``SETUP_XVFB`` because it can be useful not just for matplotlib.

README.md
```diff
@@ -92,6 +92,9 @@ Following this, various dependencies are installed depending on the following en
 * ``$DEBUG``: if `True` this turns on the shell debug mode in the install
   scripts, and provides information on the current conda install.
 
+* ``$SETUP_XVFB``: if True this makes sure e.g., interactive matplotlib
+  backends work.
```
astrofrog: add at the end of the sentence ``by starting up a X virtual framebuffer``.

astrofrog:
:+1: apart from a small comment above :)

PR #56, opened 2016-01-28
Pin matplotlib to <= 1.5.0

larrybradley:
The `photutils` sphinx test is failing again because `matplotlib 1.5.1` is being installed:
https://travis-ci.org/astropy/photutils/jobs/105487931


bsipocz:
Oops, indeed. It seems to be a mistake that the version number changed, and not just the relation. Anyway this now brakes a test, so I'll manually merge it together with a shuffling of the tests.

PR #58, opened 2016-02-01
Updating the pinning of matplotlib version for docs builds

bsipocz:
I think the current way of hard-wiring the version to <=1.5.0 for mpl for all the docs builds isn't the right way to do it. 
E.g. it will potentially cause failures for packages that require a version of 1.4.x (as it's overrides everything, and thus 1.5.0 will be installed for those packages, too even though 1.4.x will be perfectly fine for the docs build).

Also as a somewhat separate fix, this PR makes sure that SPHINX_VERSION and MATPLOTLIB_VERSION are taken into account for the docs builds, even when they are not listed as a conda dependency (thus their versions are not pinned).

@astrofrog - Could you please review it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
 
         # This is a broken test at the moment, as we manually require
         # mpl<=1.5.0 for sphinx. Once that requirement is gone this and the
-        # next two test2 can be merged together.
+        # next three tests can be merged together.
         - os: linux
           env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0 sphinx'
                MATPLOTLIB_VERSION=1.5.1 DEBUG=True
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -49,13 +49,18 @@ matrix:
 
         # This is a broken test at the moment, as we manually require
         # mpl<=1.5.0 for sphinx. Once that requirement is gone this and the
-        # next test can be merged together.
+        # next two test2 can be merged together.
```
astrofrog: test2 -> tests

astrofrog:
:+1: once the typo above is fixed - feel free to merge!

astrofrog:
Actually, maybe we can add a test that if the Matplotlib version is not specified for a sphinx build, the installed version is NOT 1.5.1?

bsipocz:
@astrofrog - Done, and test added.

astrofrog:
Thanks! Feel free to merge!

PR #59, opened 2016-02-01
Handling of pre-releases of numpy

bsipocz:
If a pre release is available install it and run the tests, otherwise abort the build (which would result a pass on travis).

It definitely works when no pre release is available, but I'm not 100% sure it would pick up the pre release as the strings may need some adjustments.  Now it's assumed "pre" is in the name of the pre-release.

@mhvk: is this what you've thought of?

### DIFF
--- a/test_env.py
+++ b/test_env.py
@@ -43,12 +43,15 @@ def test_numpy():
         os_numpy_version = os.environ['NUMPY_VERSION'].lower()
         if 'dev' in os_numpy_version:
             assert 'dev' in numpy.__version__
+        elif 'pre' in os_numpy_version:
+            assert 'pre' in numpy.__version__
         else:
             if 'stable' in os_numpy_version:
                 assert numpy.__version__.startswith(LATEST_NUMPY_STABLE)
             else:
                 assert numpy.__version__.startswith(os_numpy_version)
             assert 'dev' not in numpy.__version__
+            assert 'pre' not in numpy.__version__
 
 
 def test_astropy():
--- a/test_env.py
+++ b/test_env.py
@@ -40,18 +40,18 @@ def test_python_version():
 def test_numpy():
     if 'NUMPY_VERSION' in os.environ:
         import numpy
+        np_version = numpy.__version__
         os_numpy_version = os.environ['NUMPY_VERSION'].lower()
         if 'dev' in os_numpy_version:
-            assert 'dev' in numpy.__version__
+            assert 'dev' in np_version
         elif 'pre' in os_numpy_version:
-            assert 'pre' in numpy.__version__
+            assert re.match("[0-9.]*[0-9](a[0-9]|b[0-9]|rc[0-9])", np_version)
         else:
             if 'stable' in os_numpy_version:
-                assert numpy.__version__.startswith(LATEST_NUMPY_STABLE)
+                assert np_version.startswith(LATEST_NUMPY_STABLE)
             else:
-                assert numpy.__version__.startswith(os_numpy_version)
-            assert 'dev' not in numpy.__version__
-            assert 'pre' not in numpy.__version__
+                assert np_version.startswith(os_numpy_version)
+            assert re.match("^[0-9]+\.[0-9]+\.[0-9]", np_version)
 
 
 def test_astropy():
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -123,7 +123,7 @@ elif [[ $NUMPY_VERSION == stable ]]; then
 elif [[ $NUMPY_VERSION == pre ]]; then
     conda install $QUIET numpy
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
-    if [[ -z $(pip list -o --pre |grep numpy) ]]; then
+    if [[ -z $(pip list -o --pre | grep numpy | grep pre) ]]; then
         exit
     fi
 elif [[ ! -z $NUMPY_VERSION ]]; then
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -124,7 +124,7 @@ elif [[ $NUMPY_VERSION == pre* ]]; then
     conda install $QUIET numpy
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
     if [[ -z $(pip list -o --pre | grep numpy | \
-        grep -E "pre|[0-9]rc[0-9]|[0-9][ab][0-9]") ]]; then
+            grep -E "[0-9]rc[0-9]|[0-9][ab][0-9]") ]]; then
         exit
     fi
 elif [[ ! -z $NUMPY_VERSION ]]; then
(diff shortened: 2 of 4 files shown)


astrofrog:
@bsipocz - thanks! can we use ``NUMPY_VERSION=prerelease`` instead of ``pre`` to be more explicit?

bsipocz:
@astrofrog - Sure, it's a good point, I think we should accept both, just as we do for ``dev`` (though only advertising the explicit one). Also I'm not sure what text they are using in pre-releasing, is it "pre" or "rc" or "b", or should we just filter for all of those?

mhvk:
Yes, this is what I was looking for. Pinging @charris about the nomenclature for numpy pre-releases (or to let us know who we should ping).

charris:
@mhvk We use the sequence `1.11.0b1, 1.11.0b2, ...,1.11.0rc1, 1.11.0rc2, ..., 1.11.0` if that is what you are asking. Might also do alphas `1.11.0a1,...` at some point but haven't yet.

charris:
@rgommers Thoughts? There is no intent to use `pre` AFAIK.

bsipocz:
@charris - Thanks. Just to confirm, all of these releases are uploaded to pypi, so one can find and install them with ``pip install --pre``?

@mhvk @astrofrog - How deep do we want to go with this? I think we could test all of these, or would you prefer to stick with the ``rc`` sequence only?

charris:
@bsipocz We don't upload pre-releases to pypi yet, that is still under discussion. Let's bring in @njsmith here.

mhvk:
@bsipocz - I think we should test all pre-releases -- the earlier we find problems, the more time there is to fix them.

njsmith:
I'm in favor of uploading prereleases to pypi, exactly because it makes these kinds of things simpler. It's @charris's call as release manager ultimately, but I think I and @rgommers were both in favor in the numpy-discussion thread? (See: https://mail.scipy.org/pipermail/numpy-discussion/2016-January/074797.html)

For astropy in particular, since you depend so much on obscure little corners of the numpy api that we keep breaking, you might also consider whether you want to run a test against numpy master. The caveat is that if this breaks then you *definitely* should talk to us because it well could just be a problem on our end (well, this is true of all breakages -- we want to know! Even if the change was intentional it doesn't mean we realized ahead of time exactly what the impact would be!) But if you do want to do this then it's very easy to do now:
```
pip install --pre --upgrade --no-index --timeout=60 --trusted-host travis-dev-wheels.scipy.org -f http://travis-dev-wheels.scipy.org/ numpy
```
(That repo also has wheel builds of scipy master, in case you want to check it as well.)

bsipocz:
@njsmith: Thanks, I'll look into that url. We actually testing against numpy master, but don't do it against scipy.

@mhvk - OK, sounds good. The latest push already handles the cases listed here, not I only have to figure out the url to feed into pip.

bsipocz:
@astrofrog: do you think we can/should use the wheels @njsmith mentioned https://github.com/astropy/ci-helpers/pull/59#issuecomment-178107819 rather than building the numpy dev from the source we get from github?

mhvk:
For the builds against master, we do need something to alert us that things break. E.g., @olebole found that the current pre-release for 1.11 breaks astropy (https://github.com/astropy/astropy/issues/4548), and travis logs for recent PRs show the same failures against master. (I'll try to investigate, but it is graduate admission season here, and as I'm responsible for that, I won't have much time...)

bsipocz:
@mhvk - Good question how to notice the failing build against master, but I have no better solution at the moment, than requiring it to pass and only put it back to the allowed failures if an upstream issue is identified. Note the issue number in the travis file, and sign up some of us to that numpy issue to get notified when things are fixed. I can volunteer for this latter (to receive notifications for these issues), as they wont mix up with my other notifications as I'm not involved in numpy PRs & issues.

bsipocz:
@njsmith - Is there a link where you source wheels for the pre-releases, too?

njsmith:
@bsipocz: Not yet. And I guess it isn't worth spending a lot of time coming up with a custom solution, because soon we'll be able to use PyPI for prerelease wheels as well (see PEP 513).

test_env.py
```diff
@@ -43,12 +43,15 @@ def test_numpy():
...
             assert 'dev' in numpy.__version__
+        elif 'pre' in os_numpy_version:
+            assert 'pre' in numpy.__version__
         else:
             if 'stable' in os_numpy_version:
                 assert numpy.__version__.startswith(LATEST_NUMPY_STABLE)
             else:
                 assert numpy.__version__.startswith(os_numpy_version)
             assert 'dev' not in numpy.__version__
+            assert 'pre' not in numpy.__version__
```
rgommers: This isn't right, the version numbering you get for pre-releases is like ``x.y.z.a1/b2/rc3``. See here for a regexp that may be of use: https://github.com/numpy/numpy/blob/master/numpy/tests/test_numpy_version.py

bsipocz:
@njsmith - Sure. PyPI would be the best anyway, I was just not sure, based on the linked e-mail thread, that you already agreed to do that.

test_env.py
rgommers: And you're of course not getting a pre-release version most of the time. Only in the time frame between the first beta and the final release can there be a pre-release version on PyPi with a version number higher than the last release.

bsipocz: Thanks for the heads up and the link, I forgot to update this script. 

The idea is that we should exit the build during the install steps if there is no available pre-release compared to the latest stable, thus the test script won't run at all.

rgommers: Ah, that makes sense.

bsipocz:
Merging as this indeed seems to be working as intended with the available pre-release from pypi.

bsipocz:
@astrofrog - Do we want to support this with appveyor, too?

astrofrog:
@bsipocz - it depends how much work it would be! Ideally yes I guess?

travis/setup_dependencies_common.sh
```diff
@@ -120,6 +120,13 @@ if [[ $NUMPY_VERSION == dev* ]]; then
 elif [[ $NUMPY_VERSION == stable ]]; then
     conda install $QUIET numpy
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
+elif [[ $NUMPY_VERSION == pre* ]]; then
+    conda install $QUIET numpy
+    export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
+    if [[ -z $(pip list -o --pre | grep numpy | \
+            grep -E "[0-9]rc[0-9]|[0-9][ab][0-9]") ]]; then
+        exit
+    fi
```
astrofrog: Why is the exit needed here? Won't that stop the whole script?

bsipocz: We want to stop the script if there isn't a pre-release available/installed (as in that case it would be just another build using the ``stable`` version). 

Here is a travis run when there was no pre-release on pypi:
https://travis-ci.org/astropy/ci-helpers/jobs/106326737

And here it runs:
https://travis-ci.org/astropy/ci-helpers/jobs/112297727

astrofrog: Ah of course - could you add a comment to that effect? This part might be a pain to replicate on Windows, depending on powershell syntax.

bsipocz: Sure.

Since I'm in a good mood today I'll play a bit with powershell now, but will open an issue for the future if it seems hopeless.

PR #60, opened 2016-02-02
PIP_DEPENDENCIES should be installed after dev versions

astrofrog:
While it was correct to move the installation of dev versions of packages after ``CONDA_DEPENDENCIES``, I think they should be installed before ``PIP_DEPENDENCIES``, otherwise pip will waste time installing e.g. a stable version of astropy, only to then replace it later with the dev version. With pip, there is no risk that pip will override the dev version though (unlike conda)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,6 +43,12 @@ environment:
         NUMPY_VERSION: "1.10"
         ASTROPY_VERSION: "lts"
 
+      - PYTHON_VERSION: "3.5"
+        NUMPY_VERSION: "1.9"
+        ASTROPY_VERSION: "1.0"
+        CONDA_DEPENDENCIES: "matplotlib h5py"
+        PIP_DEPENDENCIES: "astrodendro"
+
 platform:
     -x64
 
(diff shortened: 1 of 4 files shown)


astrofrog:
I'm working on it

astrofrog:
Testing of acf2eae6499d2f1c94556600ebaac53c799523bf correctly failed on Travis (was a regression test)

appveyor.yml
```diff
@@ -41,6 +43,12 @@ environment:
         NUMPY_VERSION: "1.10"
         ASTROPY_VERSION: "lts"
 
+      - PYTHON_VERSION: "3.5"
+        NUMPY_VERSION: "stable"
```
bsipocz: I think the appveyor builds are getting rather long, and since there is only one process for the whole astropy family, it would be nice to cut back a bit. This build basically tests among others that numpy and astropy understands and works with "stable", so maybe the builds in L19 and L38 can be removed?

astrofrog: The run time is dominated by far by the development builds, so I don't think we'll gain much by removing these two, but you're right that they aren't technically needed.

bsipocz:
@astrofrog - Could you rebase it? Then it looks good to be merged.

PR #65, opened 2016-02-05
WIP: Experimental support for 32-bit builds on MacOS X

astrofrog:
For now I disabled other builds on Travis and AppVeyor.

It would be useful to have this for Astropy core (and maybe other packages)

The idea would be to have a simple non-conda mode with limited options to be able to set up simple environments in 32-bit.


bsipocz:
:+1: 

(Can I have a quick question on gitter if you are around?)

astrofrog:
Once I have this working, and before we even consider merging this, I'll open a PR to the Astropy core package to try and see if we can test using this branch of ci-helpers.

astrofrog:
@bsipocz - while I work on this, a philosophical question: this way of installing Python is not through conda, so the script is called ``setup_python_i386_osx.sh``. However, it's a bit of a pain to have to write:

```
    - if [[ $CONDA == True ]]; then
        source ci-helpers/travis/setup_conda_$TRAVIS_OS_NAME.sh;
      else
        source "ci-helpers/travis/setup_python_"$ARCH"_"$TRAVIS_OS_NAME".sh";
      fi
```

in the Travis file. What I would suggest is that we create a wrapper called ``setup.sh`` that simply delegates to the correct script, so that then users should do:

```
source ci-helpers/travis/setup.sh
```

However, we keep the existing ``setup_conda_...`` scripts, so we won't break any set-ups. We just change our docs to recommend the above script in future, and we also update it in packages in which we want to add 32-bit testing (which will require updates to ``.travis.yml`` anyway). Does that sound reasonable? If so, I'll add the wrapper script here.

bsipocz:
@astrofrog - Absolutely 100 :+1: for the ``setup.py`` wrapper.

bsipocz:
@astrofrog - Can we do (or more precisely, do we want to do) the same for linux 32bit?

astrofrog:
Closing this. I'll never have time to finish and it's not important.

PR #66, opened 2016-02-07
allow to set channels from the yml file

dpshelio:
I've just needed such option (as used in travis) for sunpy/sunpy#1629.

I'm not sure whether that's they way and I don't know how I could test this locally...

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -78,9 +78,9 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 conda config --set always_yes true
 
 if (! $env:CONDA_CHANNELS) {
-   conda config --add channels astropy-ci-extras
+   conda config --add channels astropy astropy-ci-extras
 } else {
-   conda config --add channels $conda_channels
+   conda config --add channels $env:CONDA_CHANNELS.split(" ")
 }
 # Install the build and runtime dependencies of the project.
 conda update -q conda
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -76,8 +76,12 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 
 # Conda config
 conda config --set always_yes true
-conda config --add channels astropy-ci-extras
 
+if ($conda_channels -ne $null) {
+   conda config --add channels astropy-ci-extras
+} else {
+   conda config --add channels $conda_channels
+}
 # Install the build and runtime dependencies of the project.
 conda update -q conda
 
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -78,10 +78,14 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 conda config --set always_yes true
 
 if (! $env:CONDA_CHANNELS) {
-   conda config --add channels astropy astropy-ci-extras
+   $CONDA_CHANNELS=@("astropy", "astropy-ci-extras")
 } else {
-   conda config --add channels $env:CONDA_CHANNELS.split(" ")
+   $CONDA_CHANNELS=$env:CONDA_CHANNELS.split(" ")
 }
+foreach ($CONDA_CHANNEL in $CONDA_CHANNELS) {
+   conda config --add channels $CONDA_CHANNEL
+}
+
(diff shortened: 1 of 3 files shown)


appveyor/install-miniconda.ps1
```diff
@@ -76,8 +76,12 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 
 # Conda config
 conda config --set always_yes true
-conda config --add channels astropy-ci-extras
 
+if ($conda_channels -ne $null) {
```
bsipocz: please use upper case global variable names. Also ``-ne $null`` is not needed I think, if a variable is set it should be ``True`` in the conditional (see examples below for ``ASTROPY_VERSION``).

bsipocz:
Thanks @dpshelio. I'll leave a few inclines.

It will need testing, you could add the CONDA_CHANNELS to ``appveyor.yml`` for one of the builds. 
Also please mention it in README (you can just copy over from the travis part).

appveyor/install-miniconda.ps1
bsipocz: Also it seems to me that this conditional should be the other way around, as it goes into the ``else`` part if no ``conda_channels`` was provided

bsipocz: so what about copying over the logic from ``travis/setup_dependencies_common.sh``. If it's not given set it to be equal of ``astropy-ci-extras astropy``, and then loop over and add all the channels.

dpshelio:
Thanks @bsipocz! I did indeed have the `if` logic the other way around. I've updated the PR as you suggested.

Cadair:
@dpshelio it broke. :(

appveyor/install-miniconda.ps1
```diff
@@ -76,8 +76,12 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 
 # Conda config
 conda config --set always_yes true
-conda config --add channels astropy-ci-extras
 
+if (! $env:CONDA_CHANNELS) {
+   conda config --add channels astropy astropy-ci-extras
```
dpshelio: @bsipocz - Ive also added `astropy` as a default channel as done for travis.

bsipocz: Thanks, looks good. Can you actually add multiple channels in the same command?

Cadair: I didn't think you could.

dpshelio: Ahhh! that must be the error....

Cadair: you need

```
conda config --add channels astropy --add channels astropy-ci-extras
```

bsipocz: Me neither, but I never tried it either (Tom wrote this part for travis, so I guess he had reasons to loop over the channels). We'll see in 30 mins.

bsipocz: @Cadair - But that logic wont work for the CONDA_CHANNELS variable, as we don't know how many elements are in it.

appveyor/install-miniconda.ps1
```diff
@@ -76,7 +76,15 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
...
-conda config --add channels astropy-ci-extras
+
+if (! $env:CONDA_CHANNELS) {
+   $CONDA_CHANNELS=@("astropy", "astropy-ci-extras")
+} else {
+   $CONDA_CHANNELS=$env:CONDA_CHANNELS.split(" ")
+}
+foreach ($CONDA_CHANNEL in $CONDA_CHANNELS) {
+   conda config --add $CONDA_CHANNEL
+}
```
dpshelio: all right... this should work... (I hope)

Cadair:
MOAR BUILDS! D:

bsipocz:
@Cadair - No more builds, I've cancelled the middle ones.

Cadair:
@bsipocz Winning!

dpshelio:
Why????

dpshelio:
Ok, hopefully now it runs well... (I forgot to add `channels` to the `conda config --add`)

dpshelio:
It worked!!!! Yayyy!!!

Cadair:
REJOICE! Our CI can work again!

bsipocz:
Hurray! Thanks @dpshelio!

dpshelio:
Thanks @bsipocz!!

PR #67, opened 2016-02-09
Issue with Numpy dev now that the conda Scipy/Numpy packages are compiled against MKL

astrofrog:
To reproduce (on Linux):

```
conda create -n test python=2.7 scipy cython
source activate test
pip install https://github.com/numpy/numpy/archive/master.zip
```

then run:

```python
import numpy as np
from scipy.linalg import inv

x = np.random.random((3,3))
inv(x)
```

The error is:

```
Intel MKL FATAL ERROR: Cannot load libmkl_avx.so or libmkl_def.so.
```

This repo shows a minimal failing example with ci-helpers:

https://github.com/astrofrog/debug-mkl
https://travis-ci.org/astrofrog/debug-mkl/builds/108036541

There are two possible solutions:

* As a workaround, pin Scipy to 0.15 to prevent 0.17 from getting installed
* The proper solution is probably to git clone the Numpy repository, and edit the ``site.cfg`` file to point to the conda MKL library, and make sure we use that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,10 @@ matrix:
 
         # For the Numpy dev build, we also specify a conda package that depends
         # on Numpy to make sure that Numpy dev is used (because conda will also
-        # install Numpy stable)
+        # install Numpy stable). It's also important to include scipy here to
+        # make sure that scipy routines still work correctly, because if we
+        # aren't careful, we can end up with issues with the math kernel 
+        # library.
 
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=dev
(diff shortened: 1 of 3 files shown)


astrofrog:
Checking that the regression test fails, then I'll add the fix

astrofrog:
@bsipocz - does this look ok?

bsipocz:
Looks good. What about appveyor? Technically we support the usage of dev numpy there, too. Would it be a problem?

astrofrog:
AppVeyor doesn't seem to show the issue (the regression test is passing there). I think ``nomkl`` is not available on Windows anyway: https://github.com/ContinuumIO/anaconda-issues/issues/656

bsipocz:
@astrofrog - That case please go ahead with the merge.

PR #72, opened 2016-03-05
WIP: Add Appveyor support for pre releases

bsipocz:
As said before the online virtual terminals are not real windows machines, and stdin/out and available commands are a bit ambigious, thus this needs to be tested here.


astrofrog:
@bsipocz - this might be helpful:

http://www.appveyor.com/docs/how-to/rdp-to-build-worker

You can basically connect to the actual appveyor machine until the build times out, and you have access to powershell, etc.

bsipocz:
@astrofrog - this is brilliant, thanks!

bsipocz:
@astrofrog - After failing attempts to make this work, I've found this on the appveyor forum: ``Though there is no way to terminate the build with success status.``, and thus opened this follow-up question: http://help.appveyor.com/discussions/questions/2157-terminating-a-build-with-success

What do you think the better solution? 1) leaving this PR pending -- or 2) fall back to stable and let the test run if no pre-release is available  -- until exiting in the middle of the install script without failing the whole build gets supported upstream?

I think I would slightly prefer the second solution for now, and not over advertise the feature for now :)

bsipocz:
Even if this were to work, I won't have time any time soon to put into enhancing the appveyor support here, so closing this PR.

PR #75, opened 2016-03-25
Adding the openastronomy conda channel to the default channels

bsipocz:
Maybe one or both of the others can be removed, as well?


bsipocz:
cc @mwcraig: openastronomy channel is now a default for ``ci-helpers``.

mwcraig:
Thanks!

PR #76, opened 2016-04-02
Update numpy version to latest stable version

mwcraig:
Continuum builds of numpy 1.11 have been released.

PR #78, opened 2016-04-13
Temporary fixing of sphinx version to be <1.4

bsipocz:
With the new 1.4 sphinx we run into 2 WARNINGS. This affects every package that uses astropy-helpers, thus I think it's sensible to add a requirement for the sphinx version to be ``<1.4`` here and not at the package level.
For packages that run the sphinx build without the ``-w`` option, and thus are insensitive of these warnings, there is an option to use the ``SPHINX_VERSION`` variable to override this limitation.

We can remove this once the issue is address in the helpers.

The warnings are:
```
WARNING: while setting up extension astropy_helpers.sphinx.ext.autodoc_enhancements: directive 'autoattribute' is already registered, it will be overridden
WARNING: sphinx.ext.pngmath has been deprecated. Please use sphinx.ext.imgmath instead.
```

@astrofrog - What do you think?

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -174,7 +174,7 @@ if [[ $SETUP_CMD == build_sphinx* ]] || [[ $SETUP_CMD == build_docs* ]]; then
     fi
     if [[ ! -z $SPHINX_VERSION ]]; then
         if [[ -z $(grep sphinx $pin_file) ]]; then
-            echo "matplotlib ${SPHINX_VERSION}*" >> $pin_file
+            echo "sphinx ${SPHINX_VERSION}*" >> $pin_file
         fi
     fi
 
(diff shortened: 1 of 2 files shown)


bsipocz:
(Current failures are addressed by #76, I plan to merge it first, and then rebase this.)

travis/setup_dependencies_common.sh
```diff
@@ -172,9 +172,18 @@ if [[ $SETUP_CMD == build_sphinx* ]] || [[ $SETUP_CMD == build_docs* ]]; then
...
+    # astropy-helpers. This was we allow to override this version
+    # restriction by setting this variable in .travis.yml
+    if [[ -z $SPHINX_VERSION ]]; then
+        SPHINX_VERSION='<1.4'
+    fi
+
     if [[ ! -z $SPHINX_VERSION ]]; then
         if [[ -z $(grep sphinx $pin_file) ]]; then
-            echo "matplotlib ${SPHINX_VERSION}*" >> $pin_file
+            echo "sphinx ${SPHINX_VERSION}*" >> $pin_file
```
astrofrog: Huh, was this a typo before?

astrofrog:
I agree with this solution - it's not like any of the packages are using Sphinx 1.4 features.

travis/setup_dependencies_common.sh
bsipocz: Yes, clearly no one tried to set the sphinx version before.

PR #81, opened 2016-05-12
Fixing mpl font caching issue with the suggested workaround

bsipocz:
**Do not merge until there is a new helpers release, as without https://github.com/astropy/astropy-helpers/pull/223 still brakes the py3 builds.**

It seems that there are (two ways)[https://github.com/matplotlib/matplotlib/issues/5836#issuecomment-212052820] to fix the issue we had with the mpl font caching (e.g. #56). 

The first solution is to import mpl before running the tests (sphinx build), the second is to enable caching. 
I think the first one can be sensibly done here in ci-helpers, while the second requires each package to edit their ``.travis.yml``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,27 +63,10 @@ matrix:
                DEBUG=True
                TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=="\""1.5.0"\"'
 
-        # This is a broken test at the moment, as we manually require
-        # mpl<=1.5.0 for sphinx. Once that requirement is gone this and the
-        # next four tests can be merged together.
-        - os: linux
-          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib=1.5.0 sphinx'
-               MATPLOTLIB_VERSION=1.5.1 DEBUG=True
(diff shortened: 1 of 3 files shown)


bsipocz:
Since it's worth testing this with packages before going into production, here is a travis build for phtutils. 

https://travis-ci.org/bsipocz/photutils/builds/129697065

bsipocz:
@astrofrog - This seems to be working. The py3 test failure in the photutils run (https://travis-ci.org/bsipocz/photutils/builds/129705970) is unrelated and is due astropy/ci-helpers#61. Photutils needs to be updated similarly as https://github.com/astropy/astropy/pull/4556 once we start using that helpers patch.

bsipocz:
@astrofrog - I don't understand why this build times out in ``test_regression_mkl()`` https://travis-ci.org/astropy/ci-helpers/jobs/129703321, can't see any changes in the PR that would cause it, do you?

bsipocz:
The mkl timeout disappeared, so this is good to go.

(Cancelled the appveyor, as this will need a rebase anyway once #83 is in.)

bsipocz:
@astrofrog - This is now rebased and if you think we are OK to require astropy-helpers 1.2, that this can be finally merged.

PR #82, opened 2016-05-14
Fixing mkl version to 11.3.1 for appveyor conda installs

bsipocz:
as mkl 11.3.3 has some issues with scipy (11.3.1 worked fine)

See https://github.com/astropy/astropy/pull/4907 for the start of the discussion.

[Edit]:  - There is no ``nomkl``, thus I went with hardwiring the mkl version to 11.3.1 that worked before. Everyone else is seeing this issue with scipy, and I believe this is the official conda issue for it is https://github.com/ContinuumIO/anaconda-issues/issues/792. I'll remove the version limit once it's solved upstream.

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -134,11 +134,14 @@ if ($env:CONDA_DEPENDENCIES) {
     $CONDA_DEPENDENCIES = ""
 }
 
-# Due to scipy DLL issues with mkl 11.3.3, we should not depend on mkl for now
-# as a workaround see discussion in
-# https://github.com/astropy/astropy/pull/4907#issuecomment-219200964
+# Due to scipy DLL issues with mkl 11.3.3, and as there is no nomkl option
+# for windows, we should use mkl 11.3.1 for now as a workaround see discussion
+# in https://github.com/astropy/astropy/pull/4907#issuecomment-219200964
 
-conda install -n test -q nomkl
+if ($NUMPY_OPTION -ne "") {
+   $NUMPY_OPTION_mkl = "mkl=11.3.1 " + $NUMPY_OPTION
+   $NUMPY_OPTION = $NUMPY_OPTION_mkl
+}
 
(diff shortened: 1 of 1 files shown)


bsipocz:
OK, so apparently ``nomkl`` is not available for windows now and in any foreseeable future. 

https://github.com/conda/conda-recipes/pull/586 and referenced issues within.

MSeifert04:
I don't know if that will work, locally there is no `nomkl` for windows. And it seems AppVeyor has some problems too:

```
Fetching package metadata: .......... 
Solving package specifications: .
Error:  Package missing in current win-64 channels: 
  - nomkl
 
Did you mean one of these?
 
    mkl
```
(taken from the one run appveyor test)

bsipocz:
@MSeifert04 - See my comment above :) I'll dig around a bit more, but keep this PR for the workaround.

MSeifert04:
@bsipocz Oups, sorry for the noise.

bsipocz:
@astrofrog @eteq - I expect this to pass the tests, and I'll merge it and restart the latest astropy PR builds on appveyor.

MSeifert04:
@bsipocz It seems like this can be reverted again. Locally mkl 11.3.3 build 1 works again.

bsipocz:
@MSeifert04 - Thanks for the heads up. Some people are still seeing problems with it according to the (conda issue)[ContinuumIO/anaconda-issues#792] (which is not yet closed, either), so I think it's better to wait a bit more before removing the workaround.

MSeifert04:
@bsipocz Should I open an issue so it isn't forgotten that this should just be a temporary workaround?

bsipocz:
@MSeifert04 - No need to open another issue. I get notifications from the upstream issues, and they are still open, and still a few people are seeing problems, that may or may nor be related. Also there are a few other (non windows) mkl related issues we saw at other places, so I'm happy sticking to the older version for a little longer.

Unless @astrofrog thinks otherwise, I would keep the workaround in for now.

PR #83, opened 2016-05-17
Use built-in Miniconda on AppVeyor

astrofrog:
I often see this error on AppVeyor:

https://ci.appveyor.com/project/astrofrog/glue-3d-viewer/build/1.0.265

Not sure what causes this flakiness, but it turns out Miniconda is already installed on AppVeyor, so this PR is just an experiment to see if we can just use that.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@
 environment:
 
   global:
-      PYTHON: "C:\\conda"
+      PYTHON: "C:\\Miniconda"
       MINICONDA_VERSION: "latest"
       CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\ci-helpers\\appveyor\\windows_sdk.cmd"
       PYTHON_ARCH: "64" # needs to be set for CMD_IN_ENV to succeed. If a mix
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -3,73 +3,10 @@
 # Borrwed from: Olivier Grisel and Kyle Kastner
 # License: BSD 3 clause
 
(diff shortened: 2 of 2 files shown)


bsipocz:
If this would get rid of those random fails it would be awesome. (It was quite annoying when I asap wanted to get the mkl fix out and one of the many builds always failed.)

astrofrog:
@bsipocz - there's a failure but I think it might be a real failure with the latest Numpy version. Have you seen it before?

bsipocz:
@astrofrog - No I haven't. But there is a strange thing with having different sizes for mkl that I don't understand. Versions are the same, both python and mkl wise. For any case I restarted the builds, hoping that it was just a random error.

https://ci.appveyor.com/project/Astropy/ci-helpers/build/1.0.394/job/jtx9wvyjr17ddvxm#L85

vs the last successful build:

https://ci.appveyor.com/project/Astropy/ci-helpers/build/1.0.392/job/uem58laaly7ikpgs#L96

bsipocz:
Sadly it's consistent. 

My only wild guess is that the path is not right, the buildin miniconda installs packages somewhere but doesn't update the path, thus pip doesn't see them installed.

astrofrog:
@bsipocz - I'll dig into this later today.

astrofrog:
I'm going to close this - no time to work on it and after thinking about this some more I think it's good we keep complete control over the conda installation.

bsipocz:
Today I had to compile a patched numpy for a conda env and run into the very same problem you had here with the missing mkl. The solution is to add a ``site.cfg`` file, including the path for the lib and include dirs.

Do you think that this PR would get rid of the random build fails? If yes I could give it a try based it on top of this branch.

PR #86, opened 2016-05-19
Make sure other projects don't run ci-helpers tests

astrofrog:
If we are on Travis or AppVeyor, we should check if we are running the tests in the ci-helpers repository, or whether for example someone else is running 'py.test' without arguments after having cloned ci-helpers (in which case we shouldn’t run the ci-helpers tests).

### DIFF
--- a/test_env.py
+++ b/test_env.py
@@ -4,6 +4,18 @@ import sys
 
 import pytest
 
+# If we are on Travis or AppVeyor, we should check if we are running the tests
+# in the ci-helpers repository, or whether for example someone else is running
+# 'py.test' without arguments after having cloned ci-helpers.
+
+if 'APPVEYOR_PROJECT_SLUG' in os.environ:
+    if os.environ['APPVEYOR_PROJECT_SLUG'] != 'ci-helpers':
+        pytest.skip()
+
+if 'TRAVIS_REPO_SLUG' in os.environ:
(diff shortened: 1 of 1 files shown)


bsipocz:
Oops, fair enough :)

bsipocz:
Brilliant. You run into the same problem I did in #72. Do you think we should add a mock test, e.g ``assert 1 == 1`` to foul appveyor?

astrofrog:
@bsipocz - I don't want it to skip the tests on AppVeyor though! It shouldn't be skipping them right now. I think it's because of the uppercase ``Astropy``.

bsipocz:
As I understood the issue here, is that you meant run the tests for the given package, but skip the ones in the cloned ci-helpers (aka ``test_env.py``)?

astrofrog:
@bsipocz - yes, that's correct.

bsipocz:
Ohh, now I see what caused my confusion. 
This PR fix something that we can't test for, and the fails were unrelated to the line of thoughts I had.

astrofrog:
Interestingly, py.test returns a non-zero return code if all tests are skipped, so at least we *will* be able to detect the case where e.g. the 'slug' changes and all the tests are skipped here. The fact Travis and AppVeyor pass here means the tests did not skip.

PR #90, opened 2016-05-20
Add option to force installing dependencies during egg_info builds

adrn:
This PR adds a check for an environment variable, `$FORCE_DEPENDENCIES_INSTALL`, that will force `setup_dependencies_common.sh` to build dependencies during `egg_info` builds. This is a solution to a problem that there may be other ways to solve, but has eluded me for a few months: 

I have two projects that I'm actively developing right now ("gala" and "biff"). Both contain C and Cython code along with Python code, and Biff depends on both the Cython and C code in Gala. Because of that, in order to build an egg of Biff, I need to build Gala. Locally, this is fine, but on travis my builds fail because of [this line](https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L53). With the env variable added in this PR, I can tell the setup script to build the dependencies even during an egg_info build. 

I haven't figured out how to add a test for this new functionality...

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -50,7 +50,7 @@ conda create $QUIET -n test python=$PYTHON_VERSION
 source activate test
 
 # EGG_INFO
-if [[ $SETUP_CMD == egg_info ]]; then
+if [ $SETUP_CMD == 'egg_info' ] && [ -z $FORCE_DEPENDENCIES_INSTALL ]; then
     return  # no more dependencies needed
 fi
 


bsipocz:
Thanks @adrn. Testing is easy I think, e.g. add a new build to the matrix with ``SETUP_CMD='egg_info'``, with ``FORCE_DEPENDENCIES_INSTALL=True`` and e.g. ``NUMPY_VERSION=1.11``, then test whether you have the numpy and its version, e.g. are examples in ``.travis.yml``.

travis/setup_dependencies_common.sh
```diff
@@ -50,7 +50,7 @@ conda create $QUIET -n test python=$PYTHON_VERSION
 source activate test
 
 # EGG_INFO
-if [[ $SETUP_CMD == egg_info ]]; then
+if [ $SETUP_CMD == 'egg_info' ] && [ -z $FORCE_DEPENDENCIES_INSTALL ]; then
```
bsipocz: I would prefer it to have a bool value like the rest of the similar variables (e.g. DEBUG), and test in this line for the value of it.

Also this needs a short description at the bottom of the travis section in the readme.

astrofrog:
Technically, no dependencies should be required to run ``egg_info`` though. You should be able to run ``python setup.py egg_info`` for any package in a clean environment with no other python packages, and it should succeed. If it doesn't, then pip install will fail under certain circumstances.

adrn:
@astrofrog Hmm, I guess that is an issue then. If that is the case, I don't know how to get my project to build (Biff) -- in my Cython extensions, I need to import the other project (Gala) to get include paths. See, e.g., https://github.com/adrn/biff/blob/fix-4pi/biff/setup_package.py.

adrn:
@astrofrog Oh, maybe I can just do this:

```python
    try:
        # Get gala path
        import gala
        gala_base_path = os.path.split(gala.__file__)[0]
        gala_path = os.path.join(gala_base_path, 'potential')
    except ImportError:
        gala_path = None
```

in `setup_package.py`? That would obviate my need for this env variable but not sure if there are any side effects I don't understand?

adrn:
That trick seems to have worked for my needs -- given what @astrofrog warned above about `egg_info` builds, I think this won't be necessary.

PR #96, opened 2016-06-09
Implement DEBUG environment variable for AppVeyor

astrofrog:
We need to check if it works by looking at the log of the build that uses it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,6 +32,7 @@ environment:
         CONDA_CHANNEL_PRIORITY: "True"
 
       - PYTHON_VERSION: "2.7"
+        DEBUG: "True"
 
         # For the Numpy dev build, we also specify a conda package that depends
         # on Numpy to make sure that Numpy dev is used (because conda will also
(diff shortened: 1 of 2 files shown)


bsipocz:
This is a long overdue idea, thanks for digging down for the solution.

bsipocz:
Also looking at the output there is more there than usual. However I wonder whether printing out all of the env variables would be useful as well when ``DEBUG`` is opted in. (I'm not sure how to do it, but you did that with one of your past experimental commits (that was removed after rebases)).

astrofrog:
@bsipocz - I'm not sure how to do it either! 😆

bsipocz:
@astrofrog - Maybe then just add a ``python -c "import os; print(os.environ)"`` line as debug for the env variables?
Either way please rebase this, so then we can merge it.

astrofrog:
@bsipocz - i think the last commit should do the trick. I enabled DEBUG for the simple 2.7 build. This should show the commands run and the environment variables.

bsipocz:
@astrofrog - Thanks!

bsipocz:
Thanks @astrofrog!

Issue #100, opened 2016-06-15
Introduce fall-back onto pip if conda doesn't have recent version


bsipocz:
Based on recent discussion in #97 #98 and on [gitter](https://gitter.im/astropy/astropy?at=576185c8dfb1d8aa45a3b700), we will need an option to fall back on altrnative ways of installing a package is the version available on conda is too old.

E.g. if a build asks for 'stable' astropy version (currently we have v1.2rc1), then it's not OK to install v0.3 or v0.4 or even v1.0.3 from conda, but should go instead pip install v1.1.2

astrofrog:
@bsipocz - this is done, right?

bsipocz:
Indeed!

Issue #98, opened 2016-06-15
Installing wrong version of packages


bsipocz:
This was first reported in https://github.com/astropy/astropy/pull/5094#issuecomment-226062365, the install process erroneously picked up the wrong numpy version (1.9.0b1) from the ``astropy-ci-extras`` conda channel rather than a stable released version (e.g. 1.9.3) from the default conda channel. 

There are several issues with this, first we should never pick up a beta release unless asked for it with ``NUMPY_VERSION=prerelease``. Also we should always pick up the latest available version of a given package.

As a side feature request: would be nice to find a way to specify the priority order of conda channels for cases when a package is available in several ones.

astrofrog:
I feel like astropy-ci-extras shouldn't be providing pre-releases of packages for this reason. When did we start adding these? This broke a number of builds for some other projects of mine too.

A quick workaround is to delete all pre-releases from astropy-ci-extras.

bsipocz:
I think that happened a very long time ago, hardly any updates were made to that channel. My initial workaround is to remove it from the default channels for now.

pllim:
I wish I saw this sooner! I had been scratching my head at [this failure](https://travis-ci.org/spacetelescope/mosviz/jobs/137637100) and finally had to manually pin Numpy to 1.10 to get the tests to pass.

bsipocz:
@astrofrog - It seems that we've been using ci-extras quite frequently, as for e.g. np 1.8 the latest conda astropy is  v0.4.
I'm not sure though that testing affiliated packages with v1.0.3 is any better (e.g. the latest on ``astropy-ci-helpers`` for np 1.8).

mwcraig:
I don't recall why 1.9.0b1 was added to the channel, but I ~~can delete it easily~~. Actually, don't seem to have that superpower on anaconda.org/astropy-ci-extras.

I'm guessing this is causing problems now because of the conda issue @astrofrog turned up (https://github.com/conda/conda/issues/2675).

@astrofrog or someone else with the appropriate super-powers, could you please add a repo to the astropy org on github called `conda-channel-ci-extras`? We can then hash out what matrix of versions and packages we want in that channel and get it populated. If there is a desire for any pre-release stuff we can change the label from "main" to something else (pre?) and it won't be picked up just by using the channel.

larrybradley:
It appears `travis-ci` is having issues again because `numpy 1.9.0b1` is being installed from `astropy-ci-extras`:
[...]

PR #103, opened 2016-06-20
Making "stable" version be the latest feature release

bsipocz:
This is to work around the recent conda issues when a very old astropy version was picked up when asked for "stable". We make sure that the installed version is from the latest feature release.

In most cases we don't need to force it to be the latest bugfix release, and if a package needs that bugfix they should explicitly state that version number in their travis file rather than use stable.

This is for #100 #98

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -128,8 +128,8 @@ if [[ $NUMPY_VERSION == dev* ]]; then
     # We then install Numpy itself at the bottom of this script
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
 elif [[ $NUMPY_VERSION == stable ]]; then
-    conda install $QUIET numpy
-    export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
+    conda install $QUIET numpy=$LATEST_NUMPY_STABLE
+    export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION numpy=$LATEST_NUMPY_VERSION"
 elif [[ $NUMPY_VERSION == pre* ]]; then
     conda install $QUIET numpy
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
@@ -254,17 +254,6 @@ fi
 if [[ $NUMPY_VERSION == pre* ]]; then
     $PIP_INSTALL --pre --upgrade numpy
 fi
-
-# NUMPY STABLE
-
-# Due to recent instability in conda, sometimes very old versions are picked
-# up for the "stable" numpy alias. We should revise and possibly remove this
-# once (https://github.com/conda/conda/issues/2777) is solved.
-
-if [[ $NUMPY_VERSION == stable ]]; then
-    $PIP_INSTALL --upgrade --no-deps numpy==$LATEST_NUMPY_STABLE
-fi
-
 
 # ASTROPY DEV and PRE
 
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -7,6 +7,8 @@ set -e
 conda config --set always_yes yes --set changeps1 no
 conda config --add channels defaults
 
+conda --no-channel-priority
+
 shopt -s nocasematch
 
 LATEST_ASTROPY_STABLE=1.1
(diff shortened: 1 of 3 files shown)


astrofrog:
By the way, I just opened https://github.com/conda/conda/issues/2777 because I believe there is an underlying conda bug.

travis/setup_dependencies_common.sh
```diff
@@ -9,6 +9,10 @@ conda config --add channels defaults
 
 shopt -s nocasematch
 
+LATEST_ASTROPY_STABLE=1.1
+LATEST_ASTROPY_LTS=1.0
+LATEST_NUMPY_STABLE=1.11
```
astrofrog: Do we need the last two variables here?

bsipocz: No, not yet. Depends whether we want to hack around the numpy issues ("stable" should never pick up 1.5 or 1.8), or can rely on that it will be fixed in conda soon?

bsipocz: Also it can be a plan to rename these as "minimum stable" (e.g. 1.1 for astropy and 1.10 for numpy), to allow some flexibility and only upgrade with pip when something is messed up terribly.

bsipocz:
Do we need this for appveyor, too? I haven't seen any messed up builds there, but we are not really testing for old versions anyway.

travis/setup_dependencies_common.sh
astrofrog: I don't think things will work well with Numpy, because if you update just numpy from 1.5 to 1.11, I think a lot of packages built on Numpy will fail. I really hope the conda build gets fixed soon.

bsipocz: You're right with 1.5 vs latest stable, but we do the same with when
installing the development or the prerelease version.

astrofrog: Yes - I think it usually works ok because the ABI is less likely to change from say 1.11 to dev. But strictly speaking, dev/prerelease installations could break.

bsipocz: though at the moment it may just pick up 1.5 for those, too :)

astrofrog:
@bsipocz - so for Numpy, what about the solution of checking if it's stable and if so, changing to the latest stable version, but still using conda? In pseudo-code:

```
if NUMPY_VERSION==stable: NUMPY_VERSION=LATEST_NUMPY_STABLE
```

For Astropy we can't really do that, because for e.g. Numpy 1.8 and Astropy 1.2 it might not be in conda. But the latest stable Numpy should be in conda.

astrofrog:
Conda has a new option:

```
 --no-channel-priority, --no-channel-pri, --no-chan-pri
             Package version takes precedence over channel priority (default:
             False). Note: This feature is in beta and may change in a future
             release.
```

Maybe we should be adding this to all conda commands? (in a separate PR)

travis/setup_dependencies_common.sh
```diff
@@ -266,7 +271,17 @@ if [[ $ASTROPY_VERSION == dev* ]]; then
...
+fi
+
+# ASTROPY STABLE
+
+# Due to recent instability in conda, and as new releases are not built in
+# astropy-ci-extras, this workaround ensures that we use the latest stable
+# version of astropy.
+
+if [[ $ASTROPY_VERSION == stable ]]; then
+    $PIP_INSTALL --upgrade --no-deps astropy==$LATEST_ASTROPY_STABLE
```
bsipocz: Indeed, originally meant to do it.

astrofrog:
@bsipocz - 1.2 is available on linux and mac so I think it's fine to already put 1.2 here since that code isn't used on Windows? We should make sure we add the no-channel-priority option on AppVeyor though just in case.

bsipocz:
@astrofrog - Good point.

astrofrog:
@bsipocz - so it looks like there is a config item for this:

```
conda config --set channel_priority false
```

I think it might be preferable to use that compared to adding the command-line flag?

bsipocz:
@astrofrog - Yep, I've just seen the comment on the conda issue. It turned out that my local conda where I checked for this was super old...

astrofrog:
@bsipocz - is there a way to list all available config options?

bsipocz:
My only way was to write in something invalid, then it lists the keys. Whether it is a complete list or not, I have no clue.

bsipocz:
:tada: Finally, this should fix "stuff"!

astrofrog:
Thanks @bsipocz! I've restarted some of the failing astropy-helpers PRs.

PR #107, opened 2016-06-28
Pinning conda version to 4.1.3

bsipocz:
Obscure bugs came up again in conda (see https://github.com/conda/conda/issues/2897), this time between v4.1.3 and v4.1.4 so we decided to pin the version and only upgrade when we have resources to test the behaviours of a new release.

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -87,7 +87,7 @@ conda config --set always_yes true
 conda config --add channels defaults
 
 # Install the build and runtime dependencies of the project.
-conda update -q conda=$env:CONDA_VERSION
+conda install -q conda=$env:CONDA_VERSION
 
 if (! $env:CONDA_CHANNELS) {
    $CONDA_CHANNELS=@("astropy", "openastronomy", "astropy-ci-extras")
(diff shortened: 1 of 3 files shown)


bsipocz:
I first try to get the regression and only then will push the fix.

appveyor/install-miniconda.ps1
```diff
@@ -82,6 +86,9 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 conda config --set always_yes true
 conda config --add channels defaults
 
+# Install the build and runtime dependencies of the project.
+conda update -q conda=$env:CONDA_VERSION
```
astrofrog: Any reason to not put ``install``?

bsipocz: We want to update what came in the environment? I guess. Or does ``install`` also update the dependencies? (I should really start using conda locally to get to know it better).

astrofrog: install also updates if the version doesn't match

bsipocz: dependencies, too?

astrofrog: I think so?

bsipocz:
@astrofrog - Could you please review? This is still only a workaround with pinning conda to 4.1.3, but failing with 4.1.4.

astrofrog:
👍

PR #110, opened 2016-06-29
Using previous VS in appveyor

bsipocz:
as current one gives internal compiler error for astropy dev

related issues:
https://github.com/astropy/astropy/issues/5137
https://github.com/appveyor/ci/issues/890

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -48,6 +48,8 @@ environment:
 platform:
     -x64
 
+os: Previous Visual Studio 2015
+
 install:
 
     # Install Miniconda


mwcraig:
Also now: appveyor/ci#891

bsipocz:
This passes now, so I'm merging but will keep an eye on appveyor/ci#891 for a more official workaround.

bsipocz:
Ohh, wait this doesn't solve anything for the packages, so back to square1.

mwcraig:
I think for packages we need to send an email to astropy-dev and/or the affiliated package list. Would you like the honors, or shall I? I think it makes sense to hold off until appveyor gets the Visual Studio 2015 Update 2 image going. I *think* these are the only conditions under which it will show up (unless the affiliated package itself hits a compiler issue):

+ The package is testing against astropy dev on appveyor AND on python 3.5.
+ Not testing dev on appveyor? No problem.
+ Test dev on appveyor but not 3.5? No problem.

There is another workaround that I've thought about but hesitated to suggest, which would localize the problem to astropy: we could, on every merge into master, do a build and upload to `conda-channel-ci-extras` with a `dev` label, and direct affiliated package to use that as their source for dev tests. I've been reluctant to propose it mostly because I was worried people might think it was a good idea and should be implement 😉

bsipocz:
@mwcraig - Sure, I can send out an e-mail once the new image is ready to use. 

I would not go down the route of your other workaround, it needs workarounds and logistics in several places while the original issue may only affect a very small subset of the packages (I think actually none of the affiliates test dev astropy on appveyor, except ci-helpers).

PR #93, opened 2016-05-24
Work around issue with latest VS runtime on AppVeyor with Python 3.5

astrofrog:

Issue #92, opened 2016-05-24
AppVeyor 3.5 failure


MSeifert04:
I realized that the python 3.5 appveyor build seems to fail with "Error code 1" (see for example [here](https://ci.appveyor.com/project/Astropy/ccdproc-iqpau/build/1.0.187/job/5ax1bmjpuo5p3y2r#L129))

There was a new ``msvc_runtime`` build on conda (yesterday or tonight) and it seems likely that this is somehow connected. But I cannot reproduce it locally - maybe because I'm using mkl 11.3.3 and some other packages that aren't installed in AppVeyor - so I can't investigate it.

astrofrog:
I think this could be related to https://github.com/conda-forge/staged-recipes/issues/666 - the fix is pinning the version and build of the runtime environment

astrofrog:
I'm working on a fix in ci-helpers

bsipocz:
Thanks @astrofrog for working on this!

astrofrog:
@Cadair @MSeifert04 - should be fixed, try restarting builds

bsipocz:
I've restarted the astropy and ccdproc ones.

MSeifert04:
@bsipocz @astrofrog Thank you and it seems the PRs don't fail anymore (at least not with AppVeyor :smile:) 

FWIW - It seems the corrupt package was removed from conda. Before I close this issue as resolved I wanted to ask if fixing the version is still required.

bsipocz:
@MSeifert04 - I tend to like keeping the workarounds in for a while in ci-helpers (at least until the corresponding upstream issues are closed with everyone being happy on the thread). 
I may be wrong here, but as I see our aim is to test the (astropy) packages and not whether the whole windows ecosystems works with the most cutting edge versions of everything.

MSeifert04:
@bsipocz no problem from my side. I just thought the tests should be as "cutting edge" as possible to identify upcoming problems as soon as possible. Not catching the Bugs of other packages but their features or API changes. Feel free to close the issue.

PR #114, opened 2016-07-12
Removing temporary version pinning of vs2015_runtime

bsipocz:
as the upstream issue seems to be solved.

cc @astrofrog: What do you say, can we remove this now?

(original PR was #93 to address #92)


bsipocz:
I think this is safe to do, so I'm merging this. If there is any problems coming up we can revert the changes.

PR #117, opened 2016-07-31
Added miniconda 2 or 3 option

dpshelio:
This is to avoid donwloading py2 stuff and then upgrade

### DIFF
--- a/travis/setup_conda_linux.sh
+++ b/travis/setup_conda_linux.sh
@@ -8,7 +8,12 @@ echo "==================== Starting executing ci-helpers scripts ===============
 
 # Install conda
 # http://conda.pydata.org/docs/travis.html#the-travis-yml-file
-wget https://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+if [[ "$PYTHON_VERSION" == "2.7" ]]; then
+      wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
+else
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+fi
+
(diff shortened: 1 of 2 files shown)


dpshelio:
Builds in sunpy (eg. sunpy/sunpy#1862 - [failed job](https://travis-ci.org/sunpy/sunpy/jobs/148661088)) are failing and this seems a solution that worked if using [my fork](https://travis-ci.org/dpshelio/sunpy/jobs/148677580)

bsipocz:
Sure, we can have this, but I wonder whether getting rid of the extra download overhead is actually worth the added complexity. 
This only make things better for the py3.5 tests as for py 2.6  this is gets the python3 version, and the old way would have been a plug and play even if continuum changes the default to python3, and for python 3.3 and 3.4 one still needs those version.

bsipocz:
@dpshelio - Your fails seems to be those one of those ``conda-forge`` issues @mwcraig was complaining about. E.g. you should not install python from conda-forge but from defaults.

bsipocz:
cc@astrofrog - Do you have an idea why are so many python2.7 packages picked up from ``conda-forge`` in this py3.5 build? Can we do something about it, looks very messy?
https://travis-ci.org/sunpy/sunpy/jobs/148661088#L510

dpshelio:
Yep, it was not the downloading overhead what it brought me to this patch, but the fact of anaconda installing packages of different versions of python and failing.  As you correctly said this could be a problem with `conda-forge` which it goes out of my understanding at the moment.

Regarding 2.7 condition, we could compare it numerically, using `-le` for `<=` instead of `==` so 2.6 would install `miniconda2` too.

bsipocz:
@dpshelio - Could you please restart that failing sunpy build? It worked on my fork as is, so may have been a temporary thing.

bsipocz:
🎉

bsipocz:
@dpshelio - As your problem was just fluke, do you mind if we close this?

PR #118, opened 2016-07-31
Adding defaults channel last to have higher priority

bsipocz:
To avoid picking up packages from e.g conda-forge.

E.g.:
https://travis-ci.org/bsipocz/sunpy/jobs/148693645#L495 vs. exact same build with different channel order
https://travis-ci.org/bsipocz/sunpy/jobs/148693840#L487


bsipocz:
Maybe there are cases when this is unwanted, so can put an env variable around it.

Issue #123, opened 2016-08-08
Avoid double download and install for dependencies


bsipocz:
Ideally we want to install packages in a strict order.
 1) numpy
 2) astropy
 3) dependencies

However it seems that due to the existence of the version pin file, all the dependencies are picked up during the first step. This is suboptimal for several of our typical builds (e.g. when older astropy is required, or when the build is aborted as no numpy prerelease is available, etc.).

E.g. here is a build were only numpy needed to be installed, and then the build aborted as there is no prerelease. Then in reality everything gets installed before exiting.
https://travis-ci.org/astropy/astroquery/jobs/150718956#L685

This is an LTS astropy build, yet 1.2.1 is installed first and then downgraded:
https://travis-ci.org/astropy/astroquery/jobs/150718941#L544

It's not yet clear what is an optimal workaround. To ignore the pin file, we could use e.g. `--no-pin`. Maybe adding it to the first, the numpy builds would be enough. Then most of the astropy version checking can be added to the pin file, too.

@astrofrog - Do you have a better idea or preference?

bsipocz:
This was fixed with #124

PR #124, opened 2016-08-08
Keeping order of installs

bsipocz:
This is an attempt to address #123. 
Not sure how it can be tested, but e.g. in the prerelease travis test, scipy now shouldn't be installed even though it's listed in CONDA_DEPENDENCIES.

~~If this works, I'll add more commits to deal with the astropy step.~~

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -135,20 +135,22 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
 fi
 
 # NUMPY
+# We use --no-pin to avoid installing other dependencies just yet.
+
 if [[ $NUMPY_VERSION == dev* ]]; then
     # We install nomkl here to make sure that Numpy and Scipy versions
     # installed subsequently don't depend on the MKL. If we don't do this, then
     # we run into issues when we install the developer version of Numpy
     # because it is then not compiled against the MKL, and one runs into issues
     # if Scipy *is* still compiled against the MKL.
-    conda install $QUIET nomkl
+    conda install $QUIET --no-pin nomkl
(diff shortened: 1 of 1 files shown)


travis/setup_dependencies_common.sh
```diff
@@ -135,20 +135,22 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
...
     # we run into issues when we install the developer version of Numpy
     # because it is then not compiled against the MKL, and one runs into issues
     # if Scipy *is* still compiled against the MKL.
-    conda install $QUIET nomkl
+    conda install $QUIET --no-pin nomkl
     # We then install Numpy itself at the bottom of this script
     export CONDA_INSTALL="conda install $QUIET python=$PYTHON_VERSION"
 elif [[ $NUMPY_VERSION == stable ]]; then
-    conda install $QUIET numpy=$LATEST_NUMPY_STABLE
+    conda install $QUIET --no-pin numpy=$LATEST_NUMPY_STABLE
```
astrofrog: Will it not be the case that when installing the main dependencies later, there will still be the issue that the pin file can contain ``numpy stable*``?

bsipocz: yes, probably you're right, I'll look at this again.

bsipocz: No, numpy isn't listed in the pinfile, e.g. for a ``NUMPY_VERSION=stable CONDA_DEPENDENCIES='jinja2=2.8'`` build, the pinfile contains only jinja:
```
++++cat /home/travis/miniconda/envs/test/conda-meta/pinned

jinja2  2.8*
```

bsipocz:
I'm merging this now, if there is a problem coming up we can still pin the numpy version, too.

PR #126, opened 2016-08-10
Make channel_priority configurable via environment variable

astrofrog:
I'm not sure how we can test this, apart from creating two fake test channels...

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ This does the following:
 - Register the specified channels, or if not stated the ``astropy``, ``astropy-ci-extras``, and ``openastronomy`` channels.
 - ``export PYTHONIOENCODING=UTF8``
 
-Following this, various dependencies are installed depending on the following environment variables
+Following this, various dependencies are installed depending on the following
+environment variables
 
 * ``MAIN_CMD``: if this starts with ``pycodestyle``, then the only package
   that gets installed is the ``pycodestyle`` package. Please note that the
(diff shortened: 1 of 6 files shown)


astrofrog:
I'm also removing the explicit addition of the defaults channel which I think was a temporary measure. In Glue I need to make sure we pick up the package from conda-forge when the version is equal to that in defaults.

astrofrog:
Btw I'm testing this from my fork in https://github.com/glue-viz/glue-vispy-viewers/pull/176

astrofrog:
It seems to work!

README.md
```diff
@@ -113,6 +114,9 @@ Following this, various dependencies are installed depending on the following en
   be ``SCIKIT_IMAGE_VERSION``). If specified it will override any version
   number limitations listed in ``$CONDA_DEPENDENCIES``.
 
+* ``$CONDA_CHANNEL_PRIORITY``: can be set to True or False, and affects the
```
bsipocz: Add that the default is False

astrofrog:
(note that this includes a debug commit so this should not be merged)

travis/setup_dependencies_common.sh
```diff
@@ -48,13 +47,17 @@ for channel in $CONDA_CHANNELS; do
...
 done
 
-# Make defaults channel the highest priority install packages from there
-# if available rather than from e.g. conda-forge
-conda config --add channels defaults
+if [[ -z $CONDA_CHANNEL_PRIORITY ]]; then
+    CONDA_CHANNEL_PRIORITY=false
+else
+    # Make lowercase
+    CONDA_CHANNEL_PRIORITY=`echo $CONDA_CHANNEL_PRIORITY | awk '{print tolower($0)}'`
```
bsipocz: I think we use the ``$(...)`` form for command substitution elsewhere, but this may be only as I prefer that form.

appveyor/install-miniconda.ps1
```diff
@@ -101,7 +101,7 @@ foreach ($CONDA_CHANNEL in $CONDA_CHANNELS) {
 }
 
 if (! $env:CONDA_CHANNEL_PRIORITY) {
-   $CONDA_CHANNEL_PRIORITY=false
+   $CONDA_CHANNEL_PRIORITY="false"
```
bsipocz: If it's not working, try with ``$false``? L56 uses it like that.

astrofrog:
@bsipocz - does this seem ok to you? I had to rebase, but the tests passed previously.

bsipocz:
Looks good. Thanks @astrofrog!

PR #132, opened 2016-09-08
Unfixed coverage version to use latest one

pllim:
Unfixed `coverage` version to use latest one (currently 4.2 from `conda`).

This is to possibly fix the timing problem in astropy/astropy#4826.

c/c @mhvk @MSeifert04

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -304,10 +304,8 @@ fi
 # install this now in case the user installs cpp-coveralls via PIP_DEPENDENCIES.
 
 if [[ $SETUP_CMD == *coverage* ]]; then
-    # TODO can use latest version of coverage (4.0) once astropy 1.1 is out
-    # with the fix of https://github.com/astropy/astropy/issues/4175.
     # We install requests with conda since it's required by coveralls.
-    $CONDA_INSTALL coverage==3.7.1 requests
+    $CONDA_INSTALL coverage requests
     $PIP_INSTALL coveralls
 fi
 


mhvk:
This looks good, but my one question is how to test the problems with previous 4.x versions have gone away...  Should we just merge and see if any affiliated package starts complaining?  This may be the only feasible way, but then perhaps it is a good idea to send a note to the mailing list...

Alternatively, but maybe too much work, if one can override the package (or reinstall it) in .travis.yml, we could try it in https://github.com/astropy/astropy/pull/5308

pllim:
My travis-fu is not that strong. Perhaps @astrofrog or @bsipocz can suggest a way...

bsipocz:
I recall that there were some previous problems, that's why we froze the version numbers here. So I suggest to double check it with a few mock branches for at least astropy core (and maybe a few affiliated packages like gammapy that saw the issue) before merging this. Generally we try not to experiment much in the master ``ci-helpers`` is avoidable by doing testing first.
I'll try to have a look at this later tonight.

pllim:
Thanks, @bsipocz !

bsipocz:
@pllim - Could you please rebase this? (As a safeguard PRs in ci-helpers always needs to be against the latest master, so whenever the master changes everything needs to be rebased and retested. before github allows to push the merge button.)

pllim:
@bsipocz , rebased and passed.

bsipocz:
It looks like that coverage runs with the new version, both in astropy and gammapy, so I'm happy to merge this asap.
(Although I get some failures in the test builds, I think those are unrelated and should go away once the build is run in its repository and not in my fork: https://travis-ci.org/bsipocz/astropy/builds/166497095)

@pllim  - Could you please rebase it one last time?

pllim:
@bsipocz , rebased and passed.

bsipocz:
Hurray, thanks @pllim.

PR #141, opened 2016-11-08
Use --force when uninstalling astropy

astrofrog:
Otherwise we run into issues like this:

https://travis-ci.org/astropy/regions/jobs/174034243

cc @bsipocz

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -288,8 +288,9 @@ if [[ $ASTROPY_VERSION == stable ]]; then
 
     if [[ $old_astropy == True ]]; then
         # First remove astropy from conda to make sure the version installed
-        # by pip will be used
-        conda remove astropy
+        # by pip will be used. We use --force to make sure things that depend
+        # on astropy don't cause issues or get uninstalled.
+        conda remove astropy --force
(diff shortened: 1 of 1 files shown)


bsipocz:
Looks reasonable to me.

Btw we're seeing similar issues with astroquery, e.g. https://travis-ci.org/bsipocz/astroquery/jobs/174230608 except that it wants for some reason uninstall aplpy. I guess building a new version for python 3.5.2 will be a good workaround, but it won't get rid of the underlying problem. 
If you have any idea for the solution, let me know.

astrofrog:
Isn't that just because APLpy depends on Astropy too, so it tries to uninstall all dependents? This might fix it?

bsipocz:
Oh, you're right. Apparently I looked but didn't see, focused so much on the last line.

bsipocz:
However this failure also means that the latest astropy is not picked up from ``astropy-ci-extras`` for some reason.

bsipocz:
@astrofrog - For some reason this PR didn't trigger an appveyor build (it's not even in the queue). Would you mind force pushing it to try to trigger? (Obviously that would make travis rerun again, but I haven't got a better idea)

astrofrog:
Weird - seems to be ok now though (I've pushed again)

astrofrog:
🕙

PR #142, opened 2016-11-23
Fix compatibility with pytest 3.0

astrofrog:
Pytest 3.0 no longer allows pytest.skip to be called directly - see http://doc.pytest.org/en/latest/skipping.html#imperative-xfail-from-within-a-test-or-setup-function

### DIFF
--- a/test_env.py
+++ b/test_env.py
@@ -1,8 +1,11 @@
 import os
 import re
 import sys
+from distutils.version import LooseVersion
 
 import pytest
+
+PYTEST_LT_3 = LooseVersion(pytest.__version__) < LooseVersion('3')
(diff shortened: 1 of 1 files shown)


astrofrog:
@bsipocz - ready for review! :)

bsipocz:
@astrofrog - I think I slightly preferred to have tests skipped when they are not applicable, but I can see the same amount of logic behind passing them so this all looks good to me.

astrofrog:
@bsipocz - skipping them inside the tests is no longer an option, so we'd have to add decorators to the tests. I'd be fine with that too, but don't have time to work on it :)

PR #151, opened 2017-01-08
Changing python3 version to be python 3.6 in testing

bsipocz:


bsipocz:
This should wait until matplotlib and others for python 3.6 become available on the default conda channel

astrofrog:
I don't get it - is pytest really not available for Windows/Python 3.6 yet?

bsipocz:
@astrofrog - It should be, at least ``conda search`` finds it.

astrofrog:
Can you add a temporary ``conda info pytest`` command before it's installed in the powershell script?

bsipocz:
Sure, this needed a rebase anyway (though we're still waiting for mpl to be available on default).

astrofrog:
Hmm I don't get the pytest issue on a VM. Can you print out ``conda list`` as well before the pytest install command?

astrofrog:
I wonder whether it would be worth temporarily upgrading the conda version to 4.3.4 to see if it's something fixed in more recent conda versions (4.1.3 is quite old now)

bsipocz:
@astrofrog - this is rebased on the conda version update, so hopefully appveyor won't fail with that odd issue.

astrofrog:
Ah the error is different now, just due to matplotlib. Maybe the conflict error was incorrect before.

bsipocz:
I've updated the matrix a bit as np 1.9 is not available for py3.6 and it shouldn't actually matter for us to use python3.5 for those tests. Also a note for self is that it would be great to have comments on the test builds as probably there are some superfluous ones there.

I expect one job yet to fail, LTS astropy with py36, but that should be available soon once an unrelated issue is fixed in ``conda-channel-ci-extras``.

bsipocz:
:tada: Finally a green CI.

PR #153, opened 2017-01-16
Updating conda version from 4.1.3 to 4.3.4

bsipocz:
@astrofrog - I would rather do this separately than the python 3.6. If this works, we can merge this asap and rebase the other PR that still waits for the packages to be available.

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -10,7 +10,7 @@ if (! $env:ASTROPY_LTS_VERSION) {
 }
 
 if (! $env:CONDA_VERSION) {
-   $env:CONDA_VERSION = "4.1.3"
+   $env:CONDA_VERSION = "4.3.4"
 }
 
 function DownloadMiniconda ($version, $platform_suffix) {
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -41,7 +41,7 @@ if [[ -z $PIP_DEPENDENCIES_FLAGS ]]; then
 fi
 
 if [[ -z $CONDA_VERSION ]]; then
-    CONDA_VERSION=4.1.3
+    CONDA_VERSION=4.3.4
 fi
 
 PIN_FILE_CONDA=$HOME/miniconda/conda-meta/pinned


bsipocz:
The version was pinned originally in #107. Do you think we should stick with it and keep this pinned version, or remove the pinning altogether?

I'm afraid we never added a regression test, and now can't remember whether we saw the problem ever here in ``ci-helpers`` or only in the packages that use it.

astrofrog:
@bsipocz - I think we should keep a pinned version because conda isn't the most stable package from release to release. Let's make a note (in a comment) if we ever pin it in future because of a bug.

bsipocz:
Sounds good. 

Also to easy a bit of tracking the bug workarounds I'm using the label ``workaround``. Whenever it is removed, I change it to ``solved workaround``. A few times I've already used this system to look for things to clean up.

bsipocz:
All green so merging, look out for failures are they may be due to this update.

bsipocz:
Indeed something has changed, see error here: 
https://travis-ci.org/astropy/package-template/jobs/192393463#L467

@astrofrog - do you see similar in any of your packages?

astrofrog:
This looks different to any error I've seen before but it does suggest a fix:

```
If the requested url is in fact a valid conda channel, please request that the
channel administrator create `noarch/repodata.json` and associated
`noarch/repodata.json.bz2` files, even if `noarch/repodata.json` is empty.
```

I think we might have to do that for astropy-ci-extras

bsipocz:
cc @mwcraig

astrofrog:
Seems like we should add a regression test to ci-helpers that tries to get one of the old astropy versions from astropy-ci-extras.

astrofrog:
Seems related to https://docs.continuum.io/anaconda-repository/admin/noarch-packages - I wonder if the issue is that we have no noarch packages and there is a bug in conda that causes it to crash if that's the case?

astrofrog:
Ah wait the noarch stuff is just a warning. But the error at the end is different.

bsipocz:
> Seems like we should add a regression test to ci-helpers that tries to get one of the old astropy versions from astropy-ci-extras.

Sounds reasonable, and I think we just need to do that anyway in the py3.6 branch.

astrofrog:
@bsipocz - do you want to try and follow up with the conda folk regarding the latest issue? Should we revert the changes in ci-helpers regarding the conda version? (we now know the issue is related to Matplotlib not existing, nothing to do with py.test)

bsipocz:
Sure, I'll open the issue suggested at the bottom of the error. 

We can revert the pinning, but I would rather wait a bit more for their answer and to see whether we see this problem in other packages (just merged a PR for astroquery, it should be suitable for testing this.)

bsipocz:
Ouch, looking at their issue tracker it seems that the latest releases are about fire-fighting, and are full of bugs. Testing other versions in package-template now.

mwcraig:
I wonder if we should add a build matrix entry (allowed to fail) that tries to run with with conda-canary. This is just a matter of adding `conda-canary` to the list of channels...it might help catch these things before they make it into conda releases.

Not really sure it makes sense to do in helpers, rather than in, say, astropy.

bsipocz:
@mwcraig - Probably it doesn't make sense to add here, even upgrading to 4.3.4 passed all the ci-helpers tests, and it only failed on one combination for the affiliates (it also passed everything for core).  Also looking at their issue tracker, most of their releases is full of issues, so not sure either whether it's our job to try to prevent things getting into release...

Anyway I was cc-ing you because of the warnings here: https://github.com/astropy/ci-helpers/pull/153#issuecomment-272905936. Do you think it's genuine and so we should do something about it?

mwcraig:
The latest conda added support for a noarch directory. I do not believe we are required to add one.

bsipocz:
@mwcraig - So we just need to ignore the warning? https://travis-ci.org/bsipocz/package-template/jobs/194831551#L487

mwcraig:
For now, yes. I don't really understand why conda has been set up to raise a warning in this case...one does not get a warning if, for example, you do not have a windows directory in the conda channel and I don't know why noarch is treated differently. I assume there is some good reason...

In any event, the eventual fix should be a script that adds noarch to all of the channels on anaconda.org.

PR #157, opened 2017-01-20
Unpinning conda version to the latest release 4.3.6

bsipocz:
The error we got with conda 4.3.4 in #153 got fixed for 4.3.6. A test run with package helpers passes with all green, so this should be probably OK to go in. (Another test with astropy core is still running, but I'll be able to see the results by the time the tests for this PR finish.)


bsipocz:
@astrofrog - I'm merging this as it worked both for the template and core. Please look out for weird failures though in case another corner case comes up.

PR #164, opened 2017-02-11
Updting stable numpy version to 1.12

### LINKED ISSUE #165: Hard wire stable numpy version in appveyor, too
To make travis and appveyor more similar, we need to add a hard wired latest numpy version to appveyor too (even though it was a workaround for travis originally). 
The new version 1.12 just came out on the default conda channel today, and the appveyor builds picked it up, while the travis doesn't which is an unnecessary inconsistency.

bsipocz:
This will also close #165


bsipocz:
OK, so apparently this PR needs some hacks as not all the stable combinations are available at the moment, having three major version updates so close to each other was a bit unlucky. (E.g. no astropy 1.3 for python3.6 and numpy 1.12 for windows).

bsipocz:
@astrofrog - Could you have a look at this? There's a bit of hacking with appveyor, thus I would be less worried merging this once someone else has seen it.

bsipocz:
@astrofrog - Could you have a look at this? There's a bit of hacking with appveyor, thus I would be less worried merging this once someone else has seen it.

bsipocz:
OK, numpy 1.12 is out for a while, so we better start using it. I'll keep an eye on the builds in case this PR breaks something.

astrofrog:
@bsipocz - sorry for not looking before - it looks like this just moves one of the builds but doesn't actually change anything?

astrofrog:
Looks good to me!

bsipocz:
@astrofrog - I keep moving the builds in appveyor, always having the problematic to the top, otherwise it takes ages to get debugging done.

Issue #166, opened 2017-02-15
UnsatisfiableError: with numpy=stable and astropy=stable


hamogu:
In a package build with the package template, I'm now (after numpy stable was bumped up to 1.12) seeing:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - astropy
  - numpy 1.12*
  - python 3.6*
```
That's true both with python 3.5 and 3.6:
- https://travis-ci.org/Chandra-MARX/marxs/builds/201994881
- https://travis-ci.org/Chandra-MARX/marxs/jobs/202012354

Either I'm overlooking something obvious or the required conda packages don't yet exist. In that case I suggest to roll back numpy to 1.11 for the time being (I've noticed that other affiliated packages ,e.g. photutils https://travis-ci.org/astropy/photutils/builds/202005537 seem to either pin numpy or astropy to an a specific version).

cc: @bsipocz

bsipocz:
Ouch, good point, I checked the default channel last week, but haven't done it before merging the update. Apparently conda was quite slow this time around, both the python3.6 and the numpy updates came out weeks later than the official releases.

However I would rather try to build  the new version combination in one of our channels than revert this change, also it's quite possible that conda-forge will have the update merged today.

wtbarnes:
I'm also having this `UnsatisfiableError` issue, but with 2.7, 3.4, and 3.5. The problem manifests itself a bit differently in each version (i.e. conflicts with different packages), but involves NumPy v1.12 in each case.

Travis build logs for reference:
* [2.7](https://travis-ci.org/chianti-atomic/ChiantiPy/jobs/201881354)
* [3.4](https://travis-ci.org/chianti-atomic/ChiantiPy/jobs/201881355)
* [3.5](https://travis-ci.org/chianti-atomic/ChiantiPy/jobs/201881356)

bsipocz:
@hamogu - We will definitely have a solution for the double stable case. 

@wtbarnes - I'm afraid we don't have a good solution for you. Since you've pinned the pyqt version, I would suggest to use a given, fixed numpy version for those builds as we have won't be able to provide pyqt versions for different numpy versions. Alternatively if you know that they are available on a given conda channel, add that channel to the ``CONDA_CHANNELS`` variable. Another alternative is to install pyqt via PIP_DEPENDENCIES (I'm not sure how feasible that is).

wtbarnes:
[...]

PR #167, opened 2017-02-16
Dealing with double stable cases

bsipocz:
When both astropy and numpy is set to stable

(For problems see #166)


bsipocz:
This fix brings in the issue that for being able to do ``pip install astropy==version``, we need to update the latest versions after every bugfix versions.
However this can be made as part of the release procedure ;)

bsipocz:
@astrofrog - Could you have a look at it, please? It passed travis on my fork.

bsipocz:
Double stable is working now, but other combinations still doesn't (e.g. np=stable, astropy=1.2,  https://travis-ci.org/bsipocz/astroplan/jobs/202430722)

Issue #168, opened 2017-02-20
conda install doesn't work for affiliates when pip installing astropy stable


bsipocz:
As the latest-greatest version combinations are not yet available on conda defaults, #166 introduced a workaround of pip installing astropy stable.

However this causes troubles when other astropy dependent packages are in the ``CONDA_DEPENDENCIES`` list. e.g. https://github.com/astropy/regions/issues/112#issuecomment-280928567 or https://github.com/astropy/astroplan/pull/285#issuecomment-281006421

I'm thinking about falling back pip installing in a similar way for the conda dependencies as we do now for astropy, but am a bit worried that it will break all the version pinnings, etc.

@astrofrog - do you have a better idea how to deal with this? Ideally ``conda install`` should recognize the pip installed astropy.

bsipocz:
The other way would be to trick conda resolve to find the pip installed package. For this we would need to generate  json file, which is a bigger hack than this current PR.
See https://github.com/conda/conda/issues/4701#issuecomment-281816649, and an example for such a json file https://travis-ci.org/bsipocz/ci-helpers/jobs/204445830#L638

bsipocz:
This has been fixed by #169 #170 and #174 .

PR #169, opened 2017-02-21
Falling back on pip installing

bsipocz:
This PR should be able to fix the issues in #168 by falling back on pip installing the conda dependencies (ideally only the astropy dependent ones), when astropy is installed via pip.


bsipocz:
@astrofrog - This should fix the issues (#168) we see with the affiliates that depend on other affiliates when astropy stable is pip installed. This is an ugly hack and I'm nor sure it would work in every schenario, so probably we need to have a way to opt out of it?

bsipocz:
Merging this as many affiliates are still affected by it. I'm not sure it doesn't fail in corner cases (e.g. new LTS releases, etc), but I'm certain that having this in we have more cases working than before.

PR #170, opened 2017-02-23
Failing job is the installing unsuccessful

### LINKED ISSUE #87: A failure in setting up the environment should stop the build
In the following example:

https://ci.appveyor.com/project/Eric89GXL/vispy/build/1.0.1568

There is an issue with a download, but the build proceeds anyway. This should not be the case, but not sure how to change that?

bsipocz:
This is ultimately to fix #87.
If the solution in #168  is favorable, we may want to try that for appveyor ,too (when conda install doesn't work, fall back to pip). However the exiting still need to be fixed in either case.

(I need to set up appveyor on my fork, but don't have much time atm, so letting this loose on a PR instead).

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -149,6 +149,7 @@ if ($env:ASTROPY_VERSION) {
         $ASTROPY_OPTION = "astropy=" + $env:ASTROPY_VERSION
     }
     $output = cmd /c conda install -n test -q $NUMPY_OPTION $ASTROPY_OPTION 2>&1
+    echo $output
     if ($output | select-string UnsatisfiableError) {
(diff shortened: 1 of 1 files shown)


bsipocz:
@astrofrog - Does it look good to you? The current appveyor test is showing how it will fail now before running the test script in case there the install is not working.

Compare it to the previous behavior here https://ci.appveyor.com/project/Astropy/ci-helpers/build/1.0.623/job/8lex0hv0gqpev8j5

bsipocz:
I've removed the deliberately failing job, the exiting build can be checked out here: https://ci.appveyor.com/project/Astropy/ci-helpers/build/1.0.627/job/e2jf79mkw4w1ti0i

I plan to merge this once everything passes, further adjustments can be added in a follow-up PR.

PR #171, opened 2017-02-27
Appveyor: fall back on pip install when conda install fails

bsipocz:
Basically the same logic as for travis in #169.


bsipocz:
This PR now also contains a fix for #172

bsipocz:
Since this contains a fix for a bug I'm very keen on merging asap. 

@astrofrog - Do you want to have a look before I do that?

bsipocz:
As usual, watch out for weird failures that may be due to this...

PR #174, opened 2017-03-08
Looping through conda dependencies when falling back on pip installing them

### LINKED ISSUE #173: pip install workaround brakes with non python packages
Recent workaround introduced the idea to fall back on pip install when conda install is not working, e.g. due to missing packages or incompatible version requirements. (e.g. #169)

However there are non python packages that can be conda installed, and obviously they are failing to be installed by pip (e.g. openjpeg for sunpy: https://travis-ci.org/sunpy/sunpy/jobs/207320806)

We need to add a workaround for this workaround, e.g. looping through the dependencies one by one, rather than passing the whole list from CONDA_DEPENDENCIES to pip.

bsipocz:
This PR should fix #173

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,19 @@ env:
 matrix:
     include:
 
-        # Starting with the dev versions as they take the longest to run. We
-        # deliberately test with Numpy 1.9 to check that installing Astropy dev
-        # doesn't upgrade Numpy.
+        # -> Starting with the dev versions as they take the longest to run. We
+        #    deliberately test with Numpy 1.9 to check that installing Astropy
+        #    dev doesn't upgrade Numpy.
         - os: linux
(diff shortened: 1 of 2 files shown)


Cadair:
@bsipocz I am still getting this error: https://travis-ci.org/sunpy/sunpy/jobs/209056982#L351

bsipocz:
@Cadair - could you push a  ``DEBUG=True`` to that failing job in .travis.yml?

bsipocz:
That failure is weird, I haven't got yet a good idea why it fails.

bsipocz:
@Cadair - OK, this was a tricky little beast, the matching part of the script didn't work properly and partial matches slided under the radar. The fix is coming here in ci-helpers, but in the meantime I'm opening a PR for an alternative fix for sunpy.

Issue #175, opened 2017-03-12
Test fails yet the status is green


bsipocz:
Something must have changed recently as even though there are test failures, the overall status is green here:

https://travis-ci.org/astropy/ci-helpers/jobs/210305102#L973

This may affect other astropy builds where coverage is run.

```
============================= test session starts ==============================

platform linux -- Python 3.5.3, pytest-3.0.6, py-1.4.32, pluggy-0.4.0

rootdir: /home/travis/build/astropy/ci-helpers, inifile: 

collected 10 items 

test_env.py ...F......

=================================== FAILURES ===================================

___________________________ test_dependency_imports ____________________________

    def test_dependency_imports():

    

        # We have to ignore the special case where we are running with --no-deps

        # because we don't expect that import to work.

        if os.environ.get('CONDA_DEPENDENCIES_FLAGS', '') == '--no-deps':

            return

    

        for package in dependency_list:

            if package == 'pyqt5':

                __import__('PyQt5')

            elif package == 'scikit-image':

                __import__('skimage')

            elif package == '':

                continue

            else:

>               __import__(package)

E               ImportError: No module named 'openjpeg'

test_env.py:116: ImportError

============================ pytest-warning summary ============================

WC1 None passing a string to pytest.main() is deprecated, pass a list of arguments instead.

============ 1 failed, 9 passed, 1 pytest-warnings in 0.31 seconds =============
```

bsipocz:
The tests were run by ``coverage`` rather than ``py.test``, and while coverage executes the tests it doesn't fail with failing tests. Adding a py.test run before the coverage should sufficiently cover our testing here. 

I think this issue at the end doesn't affect any other packages.

PR #176, opened 2017-03-12
Tests should run on forks, and other minor fixes

bsipocz:
This PR fixes a few minor issues.

1) tests should run on forks rather than being skipped (https://travis-ci.org/bsipocz/ci-helpers/jobs/210285831#L842)
2) coverage, or in fact any extra script should run only after success from the main test script, otherwise failures become successes (see #175)
3) openjpeg is not a python package, do not try to import it (dealt with in the 3rd commit that will be pushed after the PR is opened to test whether the fix for the 2nd point is working and the build indeed fail)
4) should also fix the issue with partial matching of package names and including/excluding them from the pin file (see https://github.com/astropy/ci-helpers/pull/174#issuecomment-286129493 and above).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,15 +45,17 @@ matrix:
           env: SETUP_CMD='test' NUMPY_VERSION=dev
                CONDA_DEPENDENCIES='scipy'
 
-        # -> Have non Python package in dependency list when falling back on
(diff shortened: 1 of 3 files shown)


bsipocz:
@Cadair - I've restarted the failing 0.7 build as am fairly certain that this fixes it.

PR #178, opened 2017-03-28
Adding license

### LINKED ISSUE #177: license of scripts
I haven't found any licences notice for this repository. I wanted to copy some scripts for another project and would like to make sure that the code is GPLv2 compatible.

bsipocz:
@astrofrog - Does it look OK to you? The license was copied over from the template.
Also I wonder whether we need the dates? If yes, is this the preferred format, or can it be open ended (2015--)?

Close #177

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -1,0 +1,25 @@
+Copyright (c) 2015-2017, Brigitta M. Sipocz & Thomas P. Robitaille
+All rights reserved.
+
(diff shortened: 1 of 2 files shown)


astrofrog:
I actually don't know if we need an end date at all (since copyright is valid for 50 years) but for consistency with what we do elsewhere, leave it as-is (with 2017 as end)

PR #182, opened 2017-04-04
fix comment string

kain88-de:
conda channels should be a space separated list. Otherwise channel addition will
fail.


bsipocz:
@kain88-de - Sadly I can't merge this as is. We use protected branches here in ci-helpers (as everything goes live once merged into master). So everything needs to be on top of the latest master and pass all CI to be identified as merge-able by github.
So either you rebase this branch, or I can cherry-pick in into a PR I'm about to open?

kain88-de:
I rebased the branch

PR #183, opened 2017-04-05
Adding pylint as a supported linter

bsipocz:


kain88-de:
It works. The time of the linting job is also halved now. Thanks

bsipocz:
@kain88-de - Yes, that's the idea for the linters to install the bare minimum at the beginning and skip the rest.

PR #199, opened 2017-06-07
Adding [docs only] and [build docs] custom tags

### LINKED ISSUE #188: Create custom skip tags
e.g ``[skip travis]`` or ``[docs only]``, etc.

related issue astropy/astropy#3933

bsipocz:
Minor clean up of [travis skip], and addition of two new custom tags to the top level script.

This address all ideas in and thus close #188

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,7 @@ env:
         - PYTHON_VERSION=3.6
 
     matrix:
-        - SETUP_CMD='egg_info'
-          TEST_CMD='python --version'
+        - SETUP_CMD='egg_info' TEST_CMD='python --version'
 
 matrix:
     include:
--- a/README.md
+++ b/README.md
@@ -35,9 +35,15 @@ This does the following:
 - Set up Miniconda.
 - Set up the PATH appropriately.
 - Set up a conda environment named 'test' and switch to it.
-- Set the ``always_yes`` config option for conda to ``true`` so that you don't need to include ``--yes``.
-- Register the specified channels, or if not stated the ``astropy``, ``astropy-ci-extras``, and ``openastronomy`` channels.
+- Set the ``always_yes`` config option for conda to ``true`` so that you
+  don't need to include ``--yes``.
+- Register the specified channels.
 - ``export PYTHONIOENCODING=UTF8``
+- Supports custom skip tags included in the commit message that are not yet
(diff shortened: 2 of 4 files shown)


bsipocz:
@Cadair @astrofrog - I fear this is something that cannot really be tested within this repo, if you up to it, please give it a go.  (E.g. without usage I've totally forgot within a month that `[skip travis]` is already a thing...)

bsipocz:
I'll send out an e-mail to astropy-dev and the maintainers list about this, but also tagging a few people who I know are keen to keep the CI footprint minimal (though it would be great to educate the less keen people about the merits of it). @mhvk @hamogu @cdeil @eteq @astrofrog @pllim @larrybradley @Cadair

All the new custom tags work like the prerelease builds, and set the commit status to pass. Obviously it *doesn't* mean that the tests are passing. We can assume that the author has good reasons for the skip, but maintainers should be aware of this, too to avoid after merge surprises.

So, now we have ``[skip travis]`` and ``[travis skip]`` to skip travis. And more importantly have ``[docs only]`` or ``[build docs]`` that will skip all jobs except the docs building ones. 

These tags cannot be tested in this repo, but I run a few tests on a fork of mine. While it works as expected, please look out for issues that are related to this.

mhvk:
Fantastic! We should be sure to document it also in the development guide.

pllim:
That's a great addition. Thanks, @bsipocz !

mwcraig:
This is excellent !!

larrybradley:
This is great.  Thanks, @bsipocz!

kain88-de:
@bsipocz I'm not sure but this might have broken the cron job for MDAnalysis. For us it is now always run. Even during a normal PR. Below is an example job with `DEBUG=True`

[Cron Job](https://travis-ci.org/MDAnalysis/mdanalysis/jobs/243968646)
[Whole Travis Build](https://travis-ci.org/MDAnalysis/mdanalysis/builds/243968634?utm_source=github_status&utm_medium=notification)

bsipocz:
@kain88-de - Indeed, this PR cleans up a bit of the logic of doing all the build checking for the need of cancelling builds before any install happens. I suggest to use common ``setup_conda.py`` (as shown in the README) rather than the OS specific setup files.

kain88-de:
Ah yeah thank you. I overlooked that we use the specific init script.

PR #212, opened 2017-06-21
Disabling channel priority for pytest install

### LINKED ISSUE #210: Preferred way to set pytest version
I've seen in the script that ci-helpers installs an older version of pytest for compatibility reasons. But we would like to use a newer version of pytest. What is the recommended way to install the most recent version? I've seen a `PYTEST_VERSION` variable is checked during the installation but I don't understand how it would make a difference in the installation since the astropy-ci-extra repository only has version 2.7.3.  Our current solution is to add pytest with a specific version to the CONDA_DEPENDENCIES variable.

bsipocz:
This is to fix #210

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,14 @@ env:
 
 matrix:
     include:
+        # -> even if CHANNEL_PRIORITY is true then the old pytest shouldn't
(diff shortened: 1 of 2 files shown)


bsipocz:
@kain88-de - Could you test whether this one indeed solves your issue in #210?

Issue #214, opened 2017-06-23
auto install pytest-cov


kain88-de:
Would it be possible to auto install pytest-cov when `--cov` is detected as an option in `SETUP_CMD`? I'm not sure if pytest-cov is currently autoinstalled but I also can't see in the script when it would be installed.

bsipocz:
We don't use that, but coverage instead. However a PR is welcome :)

PR #216, opened 2017-06-23
auto install pytest-cov

kain88-de:
If "--cov" is found in the `SETUP_CMD` variable install pytest-cov module from
conda.

fixes: #214


bsipocz:
Thanks @kain88-de!

PR #217, opened 2017-06-29
install pytest-xdist when --numprocesses is detected

### LINKED ISSUE #213: pytest-xdist has no `--parallel` option
pytest-xdist doesn't have a parallel option argument. It has a `numprocesses` argument though in the current version.  I'm not sure if `parallel` is used somewhere in the astropy ecosystem but I would suggest adding `numprocesses` to the list of recognized options that will install pytest-xdist.

Here is the full help for xdist.

```bash
distributed and subprocess testing:
  --boxed               box each test run in a separate process (unix)
  -f, --looponfail      run tests in subprocess, wait for modified files and
                        re-run failing test set until all pass.
  -n numprocesses       shortcut for '--dist=load --tx=NUM*popen', you can use
                        'auto' here for auto detection CPUs number on host
                        system
  --max-slave-restart=MAX_SLAVE_RESTART
                        maximum number of slaves that can be restarted when
                        crashed (set to zero to disable this feature)
  --dist=distmode       set mode for distributing tests to exec environments.
                        each: send each test to each available environment.
                        load: send each test to available environment.
                        (default) no: run tests inprocess, don't distribute.
  --tx=xspec            add a test execution environment. some examples: --tx
                        popen//python=python2.5 --tx socket=192.168.1.102:8888
                        --tx ssh=user@codespeak.net//chdir=testcache
  -d                    load-balance tests. shortcut for '--dist=load'
  --rsyncdir=DIR        add directory for rsyncing to remote tx nodes.
  --rsyncignore=GLOB    add expression for ignores when rsyncing to remote tx
                        nodes.
```

kain88-de:
fixes #213 

A new pytest-xdist version  has been released with the new option. This is now checking for the `numprocesses` flag and installing pytest-xdist. Unfortunately this will only work with pytest-xdist>1.18.0 so when a person is pinning a lower version it will fails. But I'm not sure how much of a problem that is in practice since pinning will cause pytest-xdist to be installed anyway.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -324,7 +324,7 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
 fi
 
 # PARALLEL BUILDS
-if [[ $SETUP_CMD == *parallel* ]]; then
+if [[ $SETUP_CMD == *parallel* || $SETUP_CMD == *numprocesses* ]]; then
     $PIP_INSTALL pytest-xdist
 fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -169,6 +169,10 @@ matrix:
                TEST_CMD='py.test -n 4 test_env.py'
 
         - os: linux
+          env: SETUP_CMD='test --numprocesses'
+               TEST_CMD='py.test --numprocesses 4 test_env.py'
+
+        - os: linux
           env: SETUP_CMD='test --open-files'
 
         - os: linux
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ environment variables
   In addition, if ``SETUP_CMD`` contains the following flags, extra dependencies are installed:
 
     * ``--coverage``: the coverage and coveralls packages are installed
-    * ``--parallel``: the pytest-xdist package is installed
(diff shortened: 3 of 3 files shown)


travis/setup_dependencies_common.sh
```diff
@@ -324,7 +324,7 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
 fi
 
 # PARALLEL BUILDS
-if [[ $SETUP_CMD == *parallel* ]]; then
+if [[ $SETUP_CMD == *parallel* || $SETUP_CMD == *--numprocesses* ]]; then
```
bsipocz: you can drop the ``--``, numprocesses as is is distinctive enough

kain88-de:
Btw I assume this should also be added to the appveyor build?

bsipocz:
@kain88-de - Originally we planned to have the same or almost the same features for appveyor, but in reality we don't really support many of them features for mostly because of lack of our time and our very limited needs for windows testing.
However if you would find it useful, please free to add as many to it as you wish.

bsipocz:
@kain88-de - Merging this now. If you're planning to work on the appveyor stuff, it can always go into another PR.

PR #224, opened 2017-08-04
Fix pytest pinning for appveyor

bsipocz:


bsipocz:
This is a super ugly hack, but as powershell is the most horrid thing out there and it doesn't listen to any reasoning...

Issue #220, opened 2017-07-17
Add "stable" and "dev" options for more packages


bsipocz:
@Cadair has asked for a ``SUNPY_VERSION`` that has similar effects as the current ``ASTROPY_VERSION``. While we can do that, I would think it would be better to move to a more generic solution and accept it for a larger number of packages. Obviously it may not be trivial due to cross dependencies, but still worth trying.

bsipocz:
Now mpl has been added, and also the foundation to add dev testing for any of the packages listed in `CONDA_DEPENDENCIES`, but we cannot be more generic as location for the dev version has to be specified somewhere. I'm considering closing this issue now?

PR #225, opened 2017-08-13
Custom version tags for sunpy

bsipocz:
This partially deals with #220. If this works, a generalization can be worked on. The only bottleneck is that the version number for the stable releases are hard-wired. If we want to generalize the understanding of "stable" and "dev" for arbitrarily packages, then we need another, more robust solution (e.g. checking pypi for the latest version, but that possibly opens up whole can of worms).

cc @Cadair @nabobalis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,13 @@ matrix:
           env: CONDA_PRIORITY=True CONDA_CHANNELS='conda-forge'
                TEST_CMD='python -c "import pytest; assert pytest.__version__.startswith(\"3.\")"'
 
+        # -> tests sunpy stable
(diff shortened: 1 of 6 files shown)


bsipocz:
@Cadair @nabobalis - please do try this out in the sunpy affiliated packages and let me know if anything is broken.

nabobalis:
@bsipocz Thanks for this. I am sure that @Cadair will put it to good use!

Cadair:
Thanks a lot @bsipocz

bsipocz:
@Cadair - at some point I'll generelize this, but will give you a heads up, so we can make sure it keeps working...

PR #233, opened 2017-08-31
ENH: Add CONDA_ENVIRONMENT support

### LINKED ISSUE #232: ENH: Add support for `environment.yml` files
Over at MNE we want to ship a Conda `environment.yml` to help people get up and running, and it would be great if we could test it with Travis. If people think it's worthwhile to add support for this to CI-helpers, I'm happy to give it a shot. Maybe a `CONDA_ENVIRONMENT` env var that, when found, just uses the given `yml` file ~~and ignores~~ EDIT: *in addition to* using the standard `CONDA_DEPENDENCIES` and `PIP_DEPENDENCIES`?

larsoner:
Closes #232.

Let's see if it works!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -188,6 +188,14 @@ matrix:
           env: SETUP_CMD='test' PIP_DEPENDENCIES='pyparsing requests'
                PIP_DEPENDENCIES_FLAGS='--log pip_install.log'
 
+        - os: linux
+          env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml'
+               TEST_CMD="python -c $'import matplotlib\ntry:\n    import scipy\nexcept ImportError:\n    pass\nelse:\n    raise RuntimeError'"
+
+        - os: linux
+          env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml' CONDA_DEPENDENCIES="scipy"
+               TEST_CMD="python -c $'import matplotlib\nimport scipy'"
+
 before_install:
     - ln -s . ci-helpers
     - source ci-helpers/travis/setup_conda.sh
--- a/README.md
+++ b/README.md
@@ -114,6 +114,11 @@ environment variables
 * ``$CONDA_CHANNELS``: this should be a space-separated string of conda
   channel names. We don't add any channel by default.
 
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -188,6 +188,9 @@ matrix:
           env: SETUP_CMD='test' PIP_DEPENDENCIES='pyparsing requests'
                PIP_DEPENDENCIES_FLAGS='--log pip_install.log'
 
+        - os: linux
+          env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml'
```
bsipocz: Please also add another job with a dummy ``CONDA_DEPENDENCIES`` or ``PIP_DEPENDENCIES`` value as well, and then can use a custom ``TEST_CMD`` to make sure they are indeed not installed.

astrofrog: Is there any reason to not allow additional dependencies to be installed on top of the environment?

larsoner: I don't see a reason to disallow it. I'd expect such behavior to be additive.

But I should probably add a test that the ones defined in the env actually do get installed.

bsipocz:
@larsoner - Would you please add a mention of this variable to the README?

.travis.yml
larsoner: Actually it looks like you understood it, I just wrote the wrong thing and apparently changed my mind while doing this PR :)

larsoner:
Travis seems to be refusing to run. Maybe 32 builds makes it angry

bsipocz:
@larsoner - don't worry, it's working. Travis is slow to pick up the builds as this stands in the same queue as every other repo on astropy (and lots of commits were made today).

larsoner:
Ahh, didn't realize `Travis` queued by org, never hit that problem before -- thanks

larsoner:
Okay comments addressed and squashed, CIs should come back happy with the latest commit. Ready for review/merge from my end.

bsipocz:
@larsoner - Yes, sadly the queue is shared. We also use the auto-cancellation feature, so the build on a commit is cancelled from queue whenever there is a new commit pushed to that branch.

bsipocz:
Tests pass and it all looks good to me. Thanks @larsoner!

larsoner:
Thanks for the quick reviews!

PR #209, opened 2017-06-19
Removing sphinx version pinning of <1.6

bsipocz:
Merge this only once astropy-helpers v2.0 is out and merged in most affiliates.

PR #234, opened 2017-09-07
Removing sphinx version limitation of <1.6

bsipocz:
A support for newer versions are released in astropy-helpers 2.0, a couple of months ago. 

If you use sphinx-gallery, it's best to require at least 0.1.11 (that was released in May, and contains a fix for the same sphinx 1.6+ issue).


bsipocz:
Oh, brilliant. I totally forgot about #209 of which this is a duplicate. However I'll close that one as CI is already running here.

Issue #229, opened 2017-08-29
Recommend cloning with --depth=1?


astrofrog:
This may not be a bad idea:

https://github.com/Anaconda-Platform/anaconda-client/blob/develop/.travis.yml#L29

Should we recommend this?

bsipocz:
Oh, sure it's very sensible idea.
 
Whenever we need to debug something from the packages point of view, we need to change the cloning anyway to point to a fork and feature branch.

pllim:
Is this still relevant? With `setuptools_scm`, shallow clone won't work anymore?

bsipocz:
We don't do versioning for ci-helpers, so I'm not sure how `setuptools_scm` changes things here?

bsipocz:
However #235 did just what was suggested, so closing this sounds reasonable (I don't think it's worth much of our time to batchPR the shallow clones downstream).

pllim:
Ah, sorry, I misunderstood the issue. Sorry for the noise!

pllim:
p.s. I agree batchpr is not necessary, as this repo is not very big anyway.

PR #235, opened 2017-09-13
DOC recommend using shallow clones in the README

rth:
Recommend using shallow clones in the README ( cf. https://github.com/astropy/ci-helpers/issues/229 ).

The difference is a 20k download size as compared to 340k for the full git history. Also reduced the run time of this step from 1s to 0.5s on my laptop...

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Include the following lines at the start of the ``install`` section in ``.travis
 
 ```yaml
 install:
-    - git clone git://github.com/astropy/ci-helpers.git
+    - git clone --depth 1 git://github.com/astropy/ci-helpers.git
     - source ci-helpers/travis/setup_conda.sh
 ```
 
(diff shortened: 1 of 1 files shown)


bsipocz:
Thank you @rth! Looks good to me.

I would suggest to remove the auto-closure ("fixes #229") from the description, as for the full fix of that issue it would be nice to mass open PRs with the suggestion to all the repos that uses ci-helpers. We have some tools in the works to automate that with a bot, and will probably sprint on them a bit next week. Keeping the issue open just makes it less likely that we forgot about it.

rth:
@bsipocz Sure, I edited the PR comment...

rth:
> it would be nice to mass open PRs with the suggestion to all the repos that uses ci-helpers

@bsipocz Off topic; do you have some infrastructure for doing that? I had attempted some prototyping of a similar functionality  [here](https://github.com/schoty/monorepo-recipes#how-does-it-work) but couldn't find time to work more on it since...

bsipocz:
@rth - @astrofrog started to work on this here https://github.com/astrofrog/batchpr

bsipocz:
Thanks @rth!

PR #237, opened 2017-09-15
ENH: Add GLX extensions and bigger screen

larsoner:
For some packages (Mayavi, vispy) having GLX is important. This is an adapted version of the line I've been using successfully to test CIs while using these packages. It should otherwise be backward compatible with what was being done before.

### DIFF
--- a/travis/setup_conda_linux.sh
+++ b/travis/setup_conda_linux.sh
@@ -11,5 +11,5 @@ source "$( dirname "${BASH_SOURCE[0]}" )"/setup_dependencies_common.sh
 
 if [[ $SETUP_XVFB == True ]]; then
     export DISPLAY=:99.0
-    sh -e /etc/init.d/xvfb start
+    /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1920x1200x24 -ac +extension GLX +render -noreset
 fi


larsoner:
(Consider this "proposal in PR form" because it was nearly as easy to open the PR as it was to make an issue to ask about opening a PR.)

bsipocz:
@larsoner - Proposal PRs are more than welcome, no need to open an issue if you're willing to address an idea. The only downside is that this will need a rebase after #236 is merged as we use as the master is a "protected" branch, meaning all PRs should be based on the latest one.

larsoner:
> the master is a "protected" branch, meaning all PRs should be based on the latest one.

That's fine, rebasing shouldn't be a problem.

Also, IIRC Travis and AppVeyor both build the `+refs/pull/XXX/merge` versions, so in theory we could also just restart the CIs as an alternative to doing a rebase.

bsipocz:
We can try that, but I'm not sure that'll be enough for the github :(

larsoner:
As in, it's actually a mode you can set in GitHub? Didn't know that. Then yes rebase is probably the only way to go :)

bsipocz:
Yes, it's supposed to be in "Settings"-->"Branches"-->"Protected branches". And then the merge button is inactive until everything required for the protection is fulfilled.
There should should be able to adjust a bit on the seriousness of protection, e.g. only require CI-s to pass, or having a review, or also be passing CI-s with up to date base. We do the latter here as if something accidentally slips through it affects many packages immediately. However it's a bit of pain if you have many PRs parallely worked on.

bsipocz:
(Ignore the appveyor issue, that's a bug being fixed in #236)

bsipocz:
@larsoner - Could you please rebase then?

larsoner:
@bsipocz done

bsipocz:
Merging this now, but everyone stay vigilant for possible failures.

bsipocz:
Thanks @larsoner!

PR #240, opened 2017-09-28
Make pip install fallback optional

### LINKED ISSUE #227: Make pip fallback optional
It would be nice to have a way to turn on/off the pip fallback mechanism

### LINKED ISSUE #239: Test using numpy=prerelease has funky log message (need better message when prerelease is not available)
Example from astropy core:
https://travis-ci.org/astropy/astropy/jobs/280665134

Example from Ginga:
https://travis-ci.org/ejeschke/ginga/jobs/280675375

Both stopped at this message:
```
Package plan for installation in environment /home/travis/miniconda/envs/test:
The following NEW packages will be INSTALLED:
    libgfortran: 3.0.0-1            
    nomkl:       1.0-0              
    numpy:       1.13.1-py36_nomkl_0 [nomkl]
    openblas:    0.2.19-0           
DEPRECATION: The default format will switch to columns in the future.
You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf
under the [list] section) to disable this warning.
```

But Travis CI reported green status. What does this mean?

bsipocz:
This is to close #227 by adding a new env variable ``PIP_FALLBACK``. By default this is True, but no pip install fallback will happen if it's set to False.

The pr also contains the comment discussed in #239, thus close #239.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,10 @@ matrix:
         - os: linux
           env: PYTHON_VERSION=3.4 PYTEST_VERSION=3.1
 
-        # Try pytest pip fallback as conda package for 3.1 is not available for py3.4 but
-        # then fail as PIP_FALLBACK is False
+        # Try pytest pip fallback as conda package for 3.1 is not available
+        # for py3.4 but then fail as PIP_FALLBACK is False. Listing this
+        # under the allowed failures as there is no way at the moment to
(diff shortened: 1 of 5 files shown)


bsipocz:
This still needs a test for PIP_FALLBACK=False, but first I wanted to see that the powershell syntax is the right one.

PR #242, opened 2017-09-30
pip fallback should not upgrade dependencies

### LINKED ISSUE #241: Pip fallback should specify --no-deps
In this build, Numpy got upgraded by the pip fallback:

https://travis-ci.org/cdeil/healpix/jobs/281350465

bsipocz:
To fix #241. pip should ideally never upgrade dependencies, or at least stick to the version numbers pinned for conda

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,11 @@ env:
 
 matrix:
     include:
+        # making sure that falling back on pip install doesn't upgrade dependencies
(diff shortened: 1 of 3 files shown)


bsipocz:
OK, apparently this doesn't make much sense for the future as ``only-if-needed`` will be the new default in pip 10+ anyway.

PR #243, opened 2017-10-03
No longer necessary to check for pinned pytest in test

drdavella:
Unless I'm mistaken, since astropy/astropy#6419 has been closed, it is no longer necessary to assert that we're using `pytest<3.2` when testing.


bsipocz:
Well, we also need to remove the limitation from the code itself, but only after we made sure that at least the project packages are OK with it.

drdavella:
@bsipocz, gotcha. This was a misunderstanding on my part; I thought that the version restriction was no longer there but this was just an oversight in the tests.

PR #244, opened 2017-10-03
Updating conda version to latest version 4.3.27

bsipocz:
We need to update to this version as the infra at the default channel changes with it.

See: https://github.com/conda/conda/issues/6069#issuecomment-333938078

Before merging we need to make sure this will work with at least astropy core and e.g. photutils/astroquery.

https://travis-ci.org/bsipocz/astropy/builds/282870686
https://travis-ci.org/bsipocz/photutils/builds/282871701
https://travis-ci.org/bsipocz/astroquery/builds/282870888

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ environment variables
   In addition, if ``SETUP_CMD`` contains the following flags, extra dependencies are installed:
 
     * ``--coverage``: the coverage and coveralls packages are installed
-    * ``--cov``: the pytest-cov and coveralls packages is installed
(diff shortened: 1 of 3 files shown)


bsipocz:
photutils and astroquery look OK. The install for astropy also looks OK, but there may be some unrelated issues with the new cython version, so I would hold this back until that's sorted out.

bsipocz:
Pinning conda version to be 4.3.21 for the failing astropy job seems to be a good workaround for now (https://travis-ci.org/bsipocz/astropy/builds/286528349), so this is good to go!

PR #247, opened 2017-10-10
coveralls should be installed the same way for --cov as for --coverage

bsipocz:
To get a bit of self consistency.

@kain88-de - Does it look OK to you? (asking as you were added the ``--cov`` trigger for the pytest-cov install)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ environment variables
   In addition, if ``SETUP_CMD`` contains the following flags, extra dependencies are installed:
 
     * ``--coverage``: the coverage and coveralls packages are installed
+    * ``--cov``: the pytest-cov and coveralls packages is installed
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -68,9 +68,9 @@ environment variables
   In addition, if ``SETUP_CMD`` contains the following flags, extra dependencies are installed:
 
     * ``--coverage``: the coverage and coveralls packages are installed
+    * ``--cov``: the pytest-cov and coveralls packages is installed
```
astrofrog: is -> are

bsipocz: Oh. I would rather fix this typo in the other PR if that's OK, so we don't have to rerun CI for this once again.

bsipocz: fixed in #244

PR #249, opened 2017-10-13
Terminate astropy lts==stable cases early

bsipocz:
No need to do it after the conda env is set up and numpy is installed.

Since the version env variables are also been moved, we shouldn't merge this until auto PRs are opened to each and every repo that uses this to switch to use ``setup_conda.py`` rather than directly the OS specific ones...

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,10 @@ environment variables
   version of Astropy. If set to ``prerelease``, the pre-release version of
   Astropy gets installed if there is any, otherwise the build exits and
   passes on Travis without running the tests. If set to ``lts`` the latest
-  long term support (LTS) version is installed (more info about LTS can be
-  found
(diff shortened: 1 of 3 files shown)


bsipocz:
Maybe don't move the numpy version out of the main file yet, just astropy and sunpy?

bsipocz:
I'm not so sure about this any more.

Now stable==2.0.x if the python version is <3.5. But I can easily imagine that a nicely fine tuned travis matrix has the LTS tested on older python while all the stable tests are run on the 3.0 branch. That case this PR would terminate the LTS test assuming it was already tested as stable.

bsipocz:
Actually this PR was only to move the termination before installing anything, so I would suggest to roll back the whole feature. What do you think @astrofrog?

PR #250, opened 2017-10-15
Add `docs only` skip instruction for Appveyor too [docs only]

matteobachetti:


bsipocz:
Oh, this is very good. Thanks!

matteobachetti:
Sadly, the hack I did doesn't work as needed, because no build its triggered and the Appveyor status does not update.
So, I'm trying a different approach now. Let's see

matteobachetti:
@bsipocz now it _should_ work. 
There's still a lot going on in the builds, installing a lot of conda packages, but it seems that solving that would complicate things a lot.

astrofrog:
In principle this would skip docs builds on appveyor, right? I don't *think* anyone is doing that at the moment, but we should make sure we keep that in mind in future.

matteobachetti:
@astrofrog you are right, I hadn't thought about this detail. If you think this might become useful, I will look into it.

bsipocz:
I'm having laptop issues, but once I have some bearable setup, can also look into it. I'm hopeful that we can utilize the initial idea of ``skip_commits`` at least in ``package-template`` in the template appveyor.yml.

matteobachetti:
@bsipocz I didn't follow up on `skip_commits` because it made the Github checks hang (just like `[skip-ci]`). I don't know if there is a workaround for that?

bsipocz:
Yes, but that's a side effect here on ci-helpers only as we have super protected master branch that requires CI to run always!

bsipocz:
I suggest to close this as the latest commit didn't work as expected, and the plans are to drop appveyor in the future in favour of travis/circleCI/azure.

matteobachetti:
Sure!

PR #252, opened 2017-10-18
update to new sunpy v0.8.2 for stable

nabobalis:
Are these all the changes I need to make?


bsipocz:
@nabobalis - Yes, thank you!

astrofrog:
@bsipocz - can we use the PyPI API to figure this out automatically instead of having to do PRs to ci-helpers?

bsipocz:
I think it's useful to wait for the conda packages to be available (at least for astropy and numpy that's the case), but I'm all for the idea to automate that one somehow.

PR #257, opened 2017-11-18
remove else clause of resolved PR 6419

kain88-de:
https://github.com/astropy/astropy/pull/6419 has been resolved and there is a mark that the else should be removed again.

I saw that when I now noticed that ci-helper don't install the latest version of pytest.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -96,10 +96,6 @@ fi
 
 if [[ ! -z $PYTEST_VERSION ]]; then
     echo "pytest ${PYTEST_VERSION}*" >> $PIN_FILE
-else
-    # Remove this else clause when
-    # https://github.com/astropy/astropy/pull/6419 is solved
-    echo "pytest <3.2" >> $PIN_FILE
 fi
 
 if [[ ! -z $PIP_VERSION ]]; then
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -82,6 +82,8 @@ source activate test
 
 # PIN FILE
 PIN_FILE=$HOME/miniconda/envs/test/conda-meta/pinned
+# ensure the PIN_FILE exists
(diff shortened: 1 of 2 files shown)


bsipocz:
Thanks @kain88-de. However I'm afraid there are still quite a few package out there that I would like to check whether they work before removing this version pinning. I'm away this coming week, once back I'll try to do those checks and come back to this PR.

In the meantime the workaround to use newer pytest is to specify the ``PYTEST_VERSION`` variable to be >3.2.

kain88-de:
Yeah I set that as well. Just thought i remove it to follow the instructions of the comment. I’m also OK to wait no hurry.

travis/setup_dependencies_common.sh
```diff
@@ -336,6 +332,8 @@ if [[ $SETUP_CMD == *build_sphinx* ]] || [[ $SETUP_CMD == *build_docs* ]]; then
     # We don't want to install everything listed in the PIN_FILE in this section
 
     for package in matplotlib sphinx; do
+        # ensure pin file exists before moving it
+        touch $PIN_FILE
```
kain88-de: It looks like this change means that the pin file isn't always present. Here I'm forcing it to be present. I didn't find anything in the manpage of `mv` to handle this more gracefully.

travis/setup_dependencies_common.sh
```diff
@@ -2,7 +2,7 @@
 
 hash -r
 
-set -e
+set -ex
```
kain88-de: I'll remove this again once you give me the OK to handle the PIN_FILE move below like I did.

travis/setup_dependencies_common.sh
bsipocz: I think it would make sense to have this at the top in L85 after the definition of ``$PIN_FILE``

kain88-de:
I already updated the test file.

bsipocz:
Indeed. Thanks.

PR #259, opened 2017-11-30
Uninstall hypothesis 

bsipocz:
as a workaround while astropy stable is <2.0.3

(an alternative would be to revert the revert of pinning pytest to <3.2, #257, this may be better as affects only the packages that rely on the astropy testing framework)

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -271,13 +271,6 @@ if [[ ! -z $ASTROPY_VERSION ]]; then
             fi))
     fi
 
-    # Force uninstall hypothesis if it's silently installed as an upstream
-    # dependency as the astropy <2.0.3 machinery is incompatible with
-    # it. But we do that here so that it will be reinstalled if it's an
-    # explicit dependency in PIP_DEPENDENCIES or CONDA_DEPENDENCIES below
-    # https://github.com/astropy/astropy/issues/6919
-    conda remove --force hypothesis || true
-
 fi
 
 # SUNPY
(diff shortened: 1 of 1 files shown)


bsipocz:
Could you @drdavella have a look, too whether this would be sufficient workaround?

travis/setup_dependencies_common.sh
```diff
@@ -230,6 +230,13 @@ fi
 
 # ASTROPY
 if [[ ! -z $ASTROPY_VERSION ]]; then
+    # force uninstall hypothesis if it's silentry installed as an upstream
```
astrofrog: silentry -> silently

travis/setup_dependencies_common.sh
```diff
@@ -230,6 +230,13 @@ fi
...
+    # dependency as the astropy <2.0.3 machinery is incompatible with
+    # it. But keep it if it's installed as an explicit dependency in
```
astrofrog: Maybe 'But keep it if it's installed' -> 'But we do this so that here so that it will be re-installed as an explicit dependency...'

bsipocz:
The issue is that it got picked up again during astropy install, so we may at the end need to be more drastic, remove at the end if it wasn't an explicit dependency and give a warning otherwise.

bsipocz:
OK, so now this passes nicely with all the tests of halotools, but if it's explicitely installed gives a warning:
https://travis-ci.org/bsipocz/halotools/jobs/310142665#L837

Obviously it's still messes up when the dependency is indirect, e.g. brought in by sunpy or astropy-healpix, etc, which should be still fine as those cases won't run their tests.

PR #267, opened 2018-01-11
remove pytest pinning on appveyor

kain88-de:
already done for travis in #257


appveyor/install-miniconda.ps1
```diff
@@ -168,9 +168,6 @@ python --version
 checkLastExitCode
 
 # CORE DEPENDENCIES
-# Remove pytest version limitation once
-# https://github.com/astropy/astropy/pull/6419 is solved
-Copy-Item ci-helpers\appveyor\pinned ${env:PYTHON}\envs\test\conda-meta\pinned
```
bsipocz: I would keep this as a placeholder (but remote the 2 comment lines above), and just remove the pytest line from that file

bsipocz:
Thanks @kain88-de! Due to master being a protected branch, this will need to be rebased to rerun the tests. Also please address the minor comment I left.

appveyor/install-miniconda.ps1
kain88-de: I added a comment that pins should be added to this file. The pytest bin is now also removed from the file. Not sure how well git treats empty files though. I never tried that.

bsipocz:
Thanks @kain88-de!

PR #276, opened 2018-02-19
Pin setuptools version to avoid Cython segmentation fault

astrofrog:
This just adds the regression test - once I see that it fails, I'll pin setuptools to 38.4 by default. Luckily, setuptools 38.5 isn't in defaults yet, but once it is all builds with Cython (e.g. astropy core) would fail without the pinning.

See https://github.com/cython/cython/issues/2104 for more details


appveyor/install-miniconda.ps1
```diff
@@ -49,6 +49,13 @@ if (! $env:CONDA_VERSION) {
    $env:CONDA_VERSION = "4.3.27"
 }
 
+# We pin the version of setuptools as setuptools 38.5.1 causes segmentation
+# faults with Cython: https://github.com/cython/cython/issues/2104. This is
+# covered by the test_cython_segfault regression test.
+if (! $env:SETUPTOOLS_VERSION) {
+   $env:SETUPTOOLS_VERSION = "38.4.0"
```
bsipocz: Please do add this to the pin files, otherwise conda may upgrade it.

astrofrog:
For some reason the pinning is not working on Travis, I need to investigate

drdavella:
Any chance this is related to the other cython-related segfault we've seen in Astropy?

astrofrog:
@drdavella - I'm pretty certain this is the same issue

astrofrog:
Well, this was so much fun, but it's basically impossible to get it to work properly, and I think the only real fix to do is https://github.com/astropy/astropy-helpers/pull/375, so closing in favour of that.

bsipocz:
@astrofrog - would this mean that packages that don't use the helpers but do use Cython will run into the segfault?

astrofrog:
No - only packages using astropy-helpers will run into this, due to the ``run_setup`` call in ``ah_bootstrap.py``.

PR #278, opened 2018-02-22
sunpy stable is now 0.8.4

Cadair:


bsipocz:
@Cadair - you need to update the test file, too. Also, feel free to remove this manual version setting and let conda decide what to pick up, I don't have any strong option that we need to keep this pinning for sunpy (unlike for numpy and astropy, given that we run into issues with those multiple times)

Cadair:
I can see the logic for keeping the pinning. We shall add this to our post-release todo list.

bsipocz:
Thanks @Cadair!

bsipocz:
ahh, this missed appveyor, but I'll fix that in another PR.

PR #285, opened 2018-02-27
check for numpy/astropy in conda-dependecy list

### LINKED ISSUE #284: Specifying numpy in CONDA_DEPENDENCIES might cause conflict
I had `CONDA_DEPENDENCIES='numpy ...'`, which caused conflict in https://travis-ci.org/spacetelescope/stginga/jobs/345804779 (xref spacetelescope/stginga#154). The conflict goes away when I remove "numpy" from that line.

I don't know if this is a bug or feature. FYI.

kain88-de:
This is a test to see if the trick works. A local test on my laptop works.

Fixes #284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -196,6 +196,9 @@ matrix:
           env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml' CONDA_DEPENDENCIES="scipy"
                TEST_CMD="python -c $'import matplotlib\nimport scipy'"
 
+        - os: linux
+          env: SETUP_CMD='test' CONDA_DEPENDENCIES='numpy'
+
     allow_failures:
         - os: linux
           env: PYTHON_VERSION=3.4 PYTEST_VERSION=3.1 PIP_FALLBACK=false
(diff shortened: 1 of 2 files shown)


kain88-de:
I found that one test install numpy in the `CONDA_DEPENDENCIES` list. This would break with such a warning change.

https://github.com/astropy/ci-helpers/blob/c644db27c44bc4828c7502327c8974580e365f44/.travis.yml#L150

Should numpy and astropy rather be filtered out from the list of dependencies instead of hard exiting the script?

The current version also seems to have a problem if there is only one package in the conda dependencies list.

kain88-de:
The current version of the code can possibly break some builds that are not configured correctly (i.e. include numpy or astropy in the conda dependencies list). Another way would be to just leave a warning in the log. That is less invasive.

This is IMHO the best option. Otherwise we would have to make a priority list of which version number to use if several options are supplied by listing numpy in the conda and or pip dependencies.

kain88-de:
This approach doesn't seem viable because an OSX test reliable crashes. I currently don't have time to look into that.

PR #287, opened 2018-03-02
Limiting sphinx not to use 1.7

bsipocz:
until https://github.com/sphinx-doc/sphinx/issues/4689 is addressed in a new release


astrofrog:
Just to be clear, I think the issue is with 1.7.1, not 1.7, but no harm in pinning ``<1.7``

PR #307, opened 2018-08-19
Simplify code when PIP_FALLBACK is False

astrofrog:
In cases where PIP_FALLBACK is False, we shouldn't capture the output from the commands (in my case this makes it harder to debug issues). The diff is simpler if you ignore whitespace changes.


bsipocz:
@astrofrog - Could you rebase and the merge when everything is green?

bsipocz:
Thanks @astrofrog!

PR #273, opened 2018-02-16
Updating conda to the 4.4.x branch

bsipocz:
This is very experimental and probably a premature upgrade, so needs to be tested with multiple packages.

@dan - could you take this branch for a spin for your packages?

https://travis-ci.org/bsipocz/astropy/builds/342440186
https://travis-ci.org/bsipocz/photutils/builds/342441641

PR #308, opened 2018-08-20
Update pinned conda version to 4.5.10

astrofrog:
This is just to see if all the tests pass with the latest version - and since it's bugfix release number 10, I'm hoping it has some degree of stability...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -155,7 +155,7 @@ matrix:
         #    can make sure the conda config was correctly updated.
         # -> Have a package dependency with alphanumerical name
         - os: linux
-          env: SETUP_CMD='test' NUMPY_VERSION=1.9 CONDA_CHANNEL_PRIORITY=True
+          env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=True
                CONDA_DEPENDENCIES='jinja2=2.8 h5py matplotlib' DEBUG=True
                ASTROPY_VERSION=1.0 CONDA_CHANNELS='astropy'
                PYTHON_VERSION='3.5'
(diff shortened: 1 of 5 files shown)


bsipocz:
Previously we tested these conda updates with some of the packages as well to avoid unpleasant surprises (astropy/photutils are the usual suspects, but it can be any non super trivial package really). 
Also, I recall to ask @drdavella to check whether that given version works well with some of his most problematic packages, but couln't find any trace of those conversations with a very quick look now.

astrofrog:
@bsipocz - you might be thinking of the conversation in #273? I've copied over one of your commits from there that appears to be needed. Once we get the CI passing here, I agree we need to test it out with a few packages.

bsipocz:
@astrofrog - Yes, that's the one, thanks for digging it up.

drdavella:
This needs to go through ASAP since we're now seeing the following test failure in various repos:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - conda==4.3.34
  - conda[version='>=4.5.11']
```

I'm not sure what's causing the error on one of the travis jobs above, though.

astrofrog:
@drdavella - there are a couple of real failures here. Do you have any time to investigate? If so, feel free to push commits to my branch!

cdeil:
In Gammapy I think all builds travis-ci builds started to fail today with the conda UnsatisfiableError that @drdavella mentioned above. See https://travis-ci.org/gammapy/gammapy/builds/424488564

eteq:
I was trying this separately in #313 (as @drdavella points out this is urgent, so hopefully this closes #312 ), but there has since been another point update, so I pushed up 4.5.11 to see if that works...

eteq:
As much as it pains me to suggest this, I wonder if it's best to merge this (since we know it at least helps *some* of the packages, which are all failing now) and try to sort out the failing test ASAP after?

astrofrog:
AppVeyor was passing before, it's just not smart enough to recognize the same commit. Ready to merge!

astrofrog:
@cdeil - can you try again to make sure this fixes things?

cdeil:
Seems to run OK again for Gammapy.

See https://github.com/gammapy/gammapy/pull/1762 or https://travis-ci.org/gammapy/gammapy/builds/424543992

Thank you all for the hotfix and for maintaining the ci-helpers!

PR #311, opened 2018-09-04
Don't install any packages with apt on Travis

astrofrog:
This is an attempt to speed up the build. I don't think we actually need the apt packages?


astrofrog:
Seems to speed up by 25%!

bsipocz:
Thanks!

PR #319, opened 2018-09-10
Miniconda version

weaverba137:
This PR allows the miniconda version to be selected programmatically, and defaults to the currently (pinned) version.

### DIFF
--- a/travis/setup_conda_linux.sh
+++ b/travis/setup_conda_linux.sh
@@ -4,7 +4,8 @@
 # Note that we pin the Miniconda version to avoid issues when new versions are released.
 # This can be updated from time to time.
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=latest
+    MINICONDA_VERSION=4.5.4
+    # MINICONDA_VERSION=latest
 fi
 wget https://repo.continuum.io/miniconda/Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh -O miniconda.sh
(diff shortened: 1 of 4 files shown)


travis/setup_conda_linux.sh
```diff
@@ -3,7 +3,11 @@
 # Install conda (http://conda.pydata.org/docs/travis.html#the-travis-yml-file)
 # Note that we pin the Miniconda version to avoid issues when new versions are released.
 # This can be updated from time to time.
-wget https://repo.continuum.io/miniconda/Miniconda3-4.5.4-Linux-x86_64.sh -O miniconda.sh
+if [[ -z "${MINICONDA_VERSION}" ]]; then
+    MINICONDA_VERSION=4.5.4
+    # MINICONDA_VERSION=latest
```
bsipocz: no need for the comments, we'll wonder over time why they've been left here.

weaverba137:
Will do...

weaverba137:
Funny enough, `$MINICONDA_VERSION` is already implemented for appveyor.

weaverba137:
So, both the Travis & Appveyor *scripts* set `MINICONDA_VERSION=4.5.4` by default.  However, the appveyor.yml file (but not the .travis.yml file) overrides this with `MINICONDA_VERSION=latest`.  Shouldn't the tests both use the same version?

bsipocz:
@weaverba137 - I think it's OK to test multiple versions, it also double checks that the machinery works.

weaverba137:
Good, thank you!

PR #322, opened 2018-09-13
export latest stable or LTS version variables

zemanj:
Export the variables
- `ASTROPY_LTS_VERSION`
- `LATEST_NUMPY_STABLE`
- `LATEST_SUNPY_STABLE`

so that these versions numbers become accessible from outside of `travis/setup_dependencies_common.sh`.

This gives some more flexibility to users who want to have custom checks requiring this information outside of ci-helpers scripts. In the AppVeyor scripts, this ~seems to be already in place~ is not easily possible (see my [comment](https://github.com/astropy/ci-helpers/pull/322#issuecomment-421365055) below).


zemanj:
I tried exporting the variables in `appveyor/install-miniconda.ps1` using commands like
```powershell
[System.Environment]::SetEnvironmentVariable("LATEST_NUMPY_STABLE", "$env:LATEST_NUMPY_STABLE", "User")
```
which successfully set the corresponding environment variables. Unfortunately, these changes are not reflected in the current session and only become effective upon restarting cmd / powershell.

Thus, I removed them again and changed the corresponding test to travis-only.

The problem is that there is no official support for refreshing environment variables in cmd / powershell. Using [tricks such as chocolatey does](https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/redirects/RefreshEnv.cmd) is IMHO not a proper solution. It would be much easier to have, e.g., a `config.ini` file where default parameters are stored. This would allow users to read or change default values. Furthermore, it might be cleaner to have these values stored in one place and not scattered accross multiple scripts...

zemanj:
[One of the Travis tests](https://travis-ci.org/astropy/ci-helpers/jobs/428621381) failed due to a `CondaHTTPError`:
```
Solving environment: failed
CondaHTTPError: HTTP 504 GATEWAY TIME-OUT for url <https://conda.anaconda.org/conda-forge/noarch/repodata.json>
Elapsed: 00:10.026206
CF-RAY: 45a337c93dff9fb4-IAD
A remote server error occurred when trying to retrieve this URL.
A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.
```

That is something that could be caught by ci-helpers, maybe triggering a retry.

zemanj:
`CondaHTTPError` seems to become kind of a show stopper. @bsipocz Could you please restart the [failed build](https://travis-ci.org/astropy/ci-helpers/jobs/428643901)? Thx!

bsipocz:
> That is something that could be caught by ci-helpers, maybe triggering a retry.

I'm not sure this is the highest priority feature to add, but if you're keen to work on it, PRs are welcome. 

Also, I see the value in factoring out the version variables but I won't have much time working on ci-helpers in the near future.

bsipocz:
Thanks @zemanj!

PR #333, opened 2018-10-25
Updating astropy versions and fix pip fallback for appveyor

bsipocz:
In addition to trivial version updates, this PR also fixes 2 serious appveyor issues.

* the python version is now pinned so conda doesn't updates e.g. 3.5 to 3.6
* pip fallback hasn't worked with pinned versions, not it has been fixed for astropy and sunpy (it's still problematic for the multi package version of `CONDA_DEPENDENCIES`, one has to be a bit more careful there as `=>` like specifications could have been there.


bsipocz:
For some reason appveyor wasn't picking this up

PR #338, opened 2018-11-05
pip fallback install to honour non exact version number

bsipocz:
This is to deal with cases when the version number is defined with a relation rather than equality


travis/setup_dependencies_common.sh
```diff
@@ -497,7 +497,11 @@ if [[ $SETUP_CMD == *build_sphinx* ]] || [[ $SETUP_CMD == *build_docs* ]]; then
             echo "Installing $package with conda was unsuccessful, using pip instead."
             PIP_PACKAGE_VERSION=$(awk '{print $2}' $PIN_FILE)
             if [[ $(echo $PIP_PACKAGE_VERSION | cut -c 1) =~ $is_number ]]; then
-                PIP_PACKAGE_VERSION='=='${PIP_${package}_VERSION}
+                if [[ ${PIP_${package}_VERSION} == *">"* || ${PIP_${package}_VERSION} == *"<"* ]]; then
```
bsipocz: this was wrong on so many levels that I'm not even sure how the test could pass. (well probably as it was a bad test). Follow-up to come.

PR #341, opened 2018-11-15
pip fallback install should really honour non exact version number

bsipocz:
This is a hopefully working approach for #338 that actually isn't working.


bsipocz:
There is a good chance that the version fall back at the `sip`, `matplotlib`, `sphinx` install section is buggy as well, but figuring that out will be beyond the scope of this PR (given that testing this whole infrastructure is quite timeconsuming, etc.)

bsipocz:
If all green, then this should be good to go.

PR #343, opened 2018-11-16
Do not terminate for stable==lts astropy any more

### LINKED ISSUE #342: LATEST_ASTROPY_VERSION vs LTS
We skip the tests when a build ask for `ASTROPY_VERSION=lts`, but the `LATEST_ASTROPY_VERSION` is also the LTS. However this logic collapses when the `LATEST_ASTROPY_VERSION` is set to LTS for python versions where the actual latest is not supported.  (e.g. currently on python 2.7 or 3.4 is someone asks for an LTS testing the build will terminate during install given that LTS is stable for that release). 

We added this logic to avoid running tests unnecessarily when the LTS and stable were the same, but I'm having regrets and seeing that there are more issues with it than perks, I suggest to remove it.

What do you think @astrofrog and @pllim?

bsipocz:
Closes #342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
         #     be picked up from the astropy-ci-extras channel.
         - os: linux
           env: CONDA_PRIORITY=True CONDA_CHANNELS='conda-forge'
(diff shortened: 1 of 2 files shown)


bsipocz:
given that for old python versions stable is set to lts, and now this would just skip those builds unnecessarily.

PR #348, opened 2018-11-27
Pin pytest to latest compatible versions if version not specified

astrofrog:
WIP - I'm going to try this out with reproject

Note that this might not be needed if we can get https://github.com/conda-forge/astropy-feedstock/pull/43 and https://github.com/conda-forge/astropy-feedstock/pull/44 merged in and propagate to the channels used here.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -188,6 +188,32 @@ if [[ $SETUP_CMD == egg_info ]]; then
     return  # no more dependencies needed
 fi
 
+# TEMPORARY PYTEST VERSION HANDLING
+
+# pytest 3.7, 3.8, and 3.9 don't work properly with astropy <3
+# (https://github.com/astropy/astropy/issues/8177)
+
+# In addition, pytest 4.0 has an issue with the test function at the top of
+# packages. There is a fix for this in the package-template
+# (https://github.com/astropy/package-template/pull/373) as well as in the core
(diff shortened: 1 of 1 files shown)


bsipocz:
@astrofrog - I've merged the conda-forge PRs, and also #349, so this will need a rebase, and probably a retest, too to see whether we need it in or not.

bsipocz:
(But actually I think we'll need it as hardly anyone uses astropy from conda-forge, so unless you also trigger the anaconda folks to rebuild we need to pin the pytest version here).

astrofrog:
I'm re-running the reproject tests (which were failing) here: https://travis-ci.org/astrofrog/reproject/builds/428950600

astrofrog:
astropy LTS builds haven't been copied over to the astropy channel yet: https://github.com/astropy/conda-channel-astropy/issues/127#issuecomment-442361539

astrofrog:
Ah right most packages will be using defaults, I thought they would be using the astropy channel (like reproject does) but I think that might be a special case.

bsipocz:
I suppose at the end we'll need this, e.g. this very recent astroplan 2.7 tests picked up pytest 4....

astrofrog:
The conda packages in defaults now specify the pytest restriction. Once this is also fixed in the astropy channel, I think we should be ok?

bsipocz:
I think we don't need this any more. 
@astrofrog - are you happy to close it?

PR #349, opened 2018-11-27
No longer pin sip

astrofrog:
This undoes https://github.com/astropy/ci-helpers/pull/299 - I *think* it's safe to do this now as I think conda packages for PyQt5 are now pinned against more specific sip versions. The pinning is actually causing issues with some packages (when trying to install glue and matplotlib for example).

(Note: In future it would be good to keep a note of the full details of the environments that e.g. segfaulted so we can verify that new conda packages are indeed available)


bsipocz:
Thanks @astrofrog. Now we need to watch out for related failures ;)

mhvk:
See #356 - this has come back... (thanks @bsipocz for making the link!)

PR #351, opened 2018-11-27
WIP: Attempt a simple solution for windows on travis

drdavella:


astrofrog:
*Attempt a simple solution* 😆

drdavella:
And there was great rejoicing 🎉 🎉 🎉 .

Now for some squash...

bsipocz:
canceled appveyor to clear the queue given travis has already failed.

drdavella:
@bsipocz I think the travis failure is actually unrelated (see #357 as well), so I was hoping this would get merged once the rest of the tests completed.

drdavella:
I'd like to try it out with ASDF but I was actually hoping this would be merged first. But I can try to use it from this branch in ASDF for now.

drdavella:
We need to account for the fact that `nomkl` is not available on Windows: https://github.com/ContinuumIO/anaconda-issues/issues/656

bsipocz:
re `nomkl`, it actually doesn't matter much, there is an empty pseudo package so even if someone asks for it, it doesn't mess up the install.

drdavella:
@bsipocz I saw a comment about that, but it appears to be not always true. I have restarted the job in question, but in my testing on `asdf` I was getting errors about no such package for `nomkl`. Unless I need to be using a different channel?

drdavella:
And just to be explicit, my `asdf` job did not pass until I made these changes here.

bsipocz:
Oh, that's good to know, thanks for the clarification.

drdavella:
Looks like we can declare success, see my test run using [asdf](https://travis-ci.org/spacetelescope/asdf/builds/463432863).

PR #357, opened 2018-12-03
Added a new mode of using ci-helpers based on pip install -e .[...]

astrofrog:
This introduces a new way of using ci-helpers where dependencies are primary installed using the dependencies defined in ``install_requires`` and ``extras_require``. This route is deliberately super minimal, and the only environment variables supported are:

* ``$PYTHON_VERSION``: used to determine the Python version to set up with conda
* ``$CONDA_DEPENDENCIES``: added to the conda create command
* ``$EXTRAS_INSTALL``: string included inside [] in the pip install command
* ``$PIP_DEPENDENCIES``: added to the pip install command

This mode is triggered by setting ``$USE_PIP_INSTALL`` to ``True``

This is still a WIP but I plan to test it out with https://github.com/astropy/astropy/pull/8198

### DIFF
--- a/travis/setup_conda_linux.sh
+++ b/travis/setup_conda_linux.sh
@@ -11,7 +11,11 @@ bash miniconda.sh -b -p $HOME/miniconda
 export PATH="$HOME/miniconda/bin:$PATH"
 
 # Install common Python dependencies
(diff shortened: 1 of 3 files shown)


bsipocz:
Looking at the downstream PR, I suppose this is now obsolete. Please reopen if you think we should still have it here.

PR #358, opened 2018-12-04
Use python -m pip instead of pip directly

drdavella:
This appears to be necessary for Windows, but it should be fine for all other platforms as well.


drdavella:
This fixed the issue I encountered when adding more Windows tests to `asdf`, so I hope it will be merged once CI passes. (See [failure](https://travis-ci.org/spacetelescope/asdf/jobs/463467499) vs [success](https://travis-ci.org/spacetelescope/asdf/jobs/463505359)).

bsipocz:
@drdavella - I'm sorry not noticing this earlier, this will now need a rebase.

PR #362, opened 2018-12-11
Updating astropy versions to 2.0.10 and 3.1

bsipocz:
Given that there is no 2.0.10 on the conda defaults, this has a slight chance of failure.

workarounds is to skip updating 2.0.9, or install 2.0.10 from the astropy channel (except that it's not yet available there).


bsipocz:
the linux versions are available, the only missing ones are osx and windows (and those were missing for 2.0.9, too)

PR #364, opened 2018-12-17
Turn errexit off before exiting commons script

### LINKED ISSUE #363: Consider unsetting bash errexit flag
This issue comes after some debugging with the Travis CI community here: https://travis-ci.community/t/build-fails-with-no-obvious-reason/1452

I've had multiple projects have failing jobs where the log shows that the job suddenly stopped (ex. https://travis-ci.org/ssec/polar2grid/jobs/468178261). As mentioned at the link above the reason the job stopped suddenly is because ci-helpers is setting `set -e` [here](https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L5). Since ci-helpers has to be sourced this setting is carried on after ci-helpers, in to the rest of the travis environment.

In my case, I had some issues in my sphinx docs `make html` (failing to find latex for building math formulas) and this was taken by travis/bash to be a failure and the script/environment exited immediately. Or at least that's what it looks like to me. I've gotten around this by doing `set +e` after sourcing ci-helpers in my `install` section, but maybe it could be done at the end of the `setup_dependencies_common.sh` script before exiting? Thoughts?

djhoese:
Closes #363 

As discussed in the related issue, setting the errexit flag in bash with `set -e` combined with ci-helpers requiring the use of `source` means that this flag is set for the rest of the travis job. In travis this causes undesired results when a user's script fails and ends up exiting the entire travis process. This leads to difficult debugging where travis doesn't even include the output for the failing command (flushed log issues maybe) and the job fails.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -748,4 +748,4 @@ if [[ ! -z $ASTROPY_VERSION ]]; then
     fi
 fi
 
-set +x
+set +ex


bsipocz:
Thank you @djhoese!

PR #368, opened 2019-01-06
MNT: Use C99 to compile numpy-dev

pllim:
As suggested by @rgommers (thank you! 🙏 ) over at https://github.com/numpy/numpy/pull/12610#issuecomment-451721733 

I will open an experimental PR over as Astropy to use this branch and see what happens. 🤞 

Fix astropy/astropy#8318

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -327,6 +327,8 @@ if [[ ! -z $(echo $CONDA_DEPENDENCIES | grep '\bmkl\b') ||
 fi
 
 if [[ $NUMPY_VERSION == dev* ]]; then
+    # We use C99 to build Numpy
+    export CFLAGS="-std=c99"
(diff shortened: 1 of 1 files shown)


pllim:
Unfortunately, this does not seem to work. Example log: https://travis-ci.org/astropy/astropy/jobs/475919446 (and also the CI for this PR has the same error)

`option -s not recognized`

rgommers:
Try `export CFLAGS="-std=c99"`, if that also doesn't work then I need some sleep first:)

pllim:
Yay, tests passed! (Appveyor failure is unrelated.)

pllim:
I guess we can still nitpick on whether we want that CFLAGS defined for all Numpy selections, but at least the solution is at hand now.

bsipocz:
Thanks @pllim and @rgommers!

PR #371, opened 2019-01-06
EXP: Disable SSL verification for Appveyor CI

### LINKED ISSUE #369: TST: Appveyor CI is broken
Appears to have some sort of HTTP error in the first job (PY2), causing the rest to be cancelled. The CI history shows same error for different PRs, suggesting that this is a real problem with Appveyor CI for this repo, not caused by the individually unrelated PRs.

pllim:
Attempt to fix #369 . Even if this works, this should be a last resort kinda solution, as warned in https://conda.io/docs/user-guide/configuration/disable-ssl-verification.html . This is one of the things suggested over at conda/conda#6007 , which helped some but not others. 🤷‍♀️

### DIFF
--- a/appveyor/install-miniconda.ps1
+++ b/appveyor/install-miniconda.ps1
@@ -240,6 +240,9 @@ $env:PATH = "${env:PYTHON};${env:PYTHON}\Scripts;" + $env:PATH
 
 # Conda config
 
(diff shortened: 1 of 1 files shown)


pllim:
Nope.

PR #373, opened 2019-01-06
EXP: Move Appveyor PY2 job down a little

pllim:
Just to see if this changes anything for #369. Whee...


pllim:
Does not change anything but at least the log is a bit cleaner...

PR #376, opened 2019-01-06
MNT: Only set CFLAGS if not already set

pllim:
Follow up of #368. This only sets `CFLAGS` in a numpy-dev job if it is not already set. Otherwise, it would overwrite calling package's setting. If the calling package sets its own `CFLAGS`, then it is on them to set it properly for a numpy-dev job themselves.

p.s. Please double check the BASH syntax. Thanks!


bsipocz:
@pllim - Could you please rebase?

pllim:
Rebased. 🤞

bsipocz:
Thanks @pllim!

PR #375, opened 2019-01-06
EXP: Update default miniconda version

pllim:
Yet another shot in the dark for #369.

PR #378, opened 2019-01-07
Bump default miniconda version

pllim:
#375 again


bsipocz:
OK, let's try this. I've added another commit to bump the version for linux and osx, too.

We need to be vigilant in the next couple of days, and revert this if packages run into issues.

pllim:
Huh, only Appveyor and the Windows build in Travis passed. Linux and OSX both got stuck at "installation finished" and failed.

bsipocz:
Thanks @pllim!

PR #379, opened 2019-01-08
Install codecov in addition to coveralls

astrofrog:
Since some packages are now using codecov, it makes sense to provide it here


bsipocz:
Thanks @astrofrog!

PR #381, opened 2019-01-15
Reverting #287

bsipocz:
Since many sphinx versions have been released since 1.7.1


bsipocz:
Locally both astroquery and photutils are happy with the latest 1.8.3 release, so we should merge this once CI is all green.

PR #383, opened 2019-01-23
overridable env variable MACOSX_DEPLOYMENT_TARGET=10.9

zemanj:
Set default OSX deployment target version to 10.9 since this is required for compiling C++ code with llvm/clang when `-stdlib=libc++` is specified.

This is accomplished with the environment variable `MACOSX_DEPLOYMENT_TARGET` which is set to `10.9` if previously empty, unset if previously set to `"clang_default"`, or, otherwise, left as is.

See also issue https://github.com/MDAnalysis/mdanalysis/issues/2180
or PR https://github.com/MDAnalysis/mdanalysis/pull/2184

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,11 @@ environment variables
 * ``RETRY_DELAY``: a positive integer specifying the number of seconds to wait
   before retrying. If not set, this will default to ``RETRY_DELAY=2``.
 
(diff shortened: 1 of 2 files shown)


bsipocz:
Thanks @zemanj!

bsipocz:
The test failures are unrelated as it seems the sunpy guys have messed up something for their dev build. So I'm going ahead and merge this now.

Thanks @zemanj!

PR #388, opened 2019-03-06
pip fallback if numpy build for specified python version is missing

zemanj:
### Problem ###
Since #382, builds with `PYTHON_VERSION=3.5` and `NUMPY_VERSION=stable` lead to Python 3.7 being installed because conda-forge doesn't provide a py35 build of numpy 1.16.
This leads either to
* tests with a wrong Python version, or
* build failure if packages without a pip candidate are required.

See also [MDAnalysis Issue #2217](https://github.com/MDAnalysis/mdanalysis/issues/2217).

### Solution ###
* Do not change python version if no matching numpy build available, i.e. add `PYTHON_OPTION` to `conda install` call during numpy installation.
* If installing numpy with conda fails, remove `NUMPY_OPTION` from `CONDA_INSTALL` and add it to `PIP_DEPENDENCIES` instead.

I currently don't have a Windows machine around to tinker with the PS script. So, for now, this PR just fixes the issue for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -190,6 +190,9 @@ matrix:
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
                CONDA_DEPENDENCIES_FLAGS='--no-deps'
+               NUMPY_VERSION=stable PYTHON_VERSION=3.5
+               CONDA_CHANNELS='conda-forge'
+               
 
         - os: linux
           env: SETUP_CMD='test' PIP_DEPENDENCIES='pyparsing requests'
--- a/.travis.yml
+++ b/.travis.yml
@@ -190,9 +190,12 @@ matrix:
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
                CONDA_DEPENDENCIES_FLAGS='--no-deps'
-               NUMPY_VERSION=stable PYTHON_VERSION=3.5
-               CONDA_CHANNELS='conda-forge'
-               
+               NUMPY_VERSION=1.16 PYTHON_VERSION=3.5
+
+        - os: linux
+          env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
+               CONDA_DEPENDENCIES_FLAGS='--no-deps'
+               NUMPY_VERSION=1.16.1 PYTHON_VERSION=3.5
 
         - os: linux
           env: SETUP_CMD='test' PIP_DEPENDENCIES='pyparsing requests'
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -112,6 +112,7 @@ export LATEST_SUNPY_STABLE=0.9.2
 
 is_number='[0-9]'
 is_eq_number='=[0-9]'
+is_eq_float="=[0-9]+\.[0-9]+"
 
 
 if [[ -z $PIP_FALLBACK ]]; then
@@ -326,6 +327,8 @@ if [[ ! -z $(echo $CONDA_DEPENDENCIES | grep '\bmkl\b') ||
     MKL=''
 fi
 
+# determine how to install numpy:
+NUMPY_INSTALL=''
 if [[ $NUMPY_VERSION == dev* ]]; then
     # We use C99 to build Numpy.
     # If CFLAGS already defined by calling pkg, it's up to them to set this.
@@ -341,13 +344,13 @@ if [[ $NUMPY_VERSION == dev* ]]; then
     # We then install Numpy itself at the bottom of this script
     export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $MKL"
 elif [[ $NUMPY_VERSION == stable ]]; then
-    retry_on_known_error conda install $QUIET --no-pin numpy=$LATEST_NUMPY_STABLE $MKL
     export NUMPY_OPTION="numpy=$LATEST_NUMPY_STABLE"
-    export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION numpy=$LATEST_NUMPY_STABLE $MKL"
+    export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $NUMPY_OPTION $MKL"
+    NUMPY_INSTALL="conda install $QUIET --no-pin $PYTHON_OPTION $NUMPY_OPTION $MKL"
 elif [[ $NUMPY_VERSION == pre* ]]; then
-    retry_on_known_error conda install $QUIET --no-pin $MKL numpy
     export NUMPY_OPTION=""
     export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $MKL"
+    NUMPY_INSTALL="conda install $QUIET --no-pin $PYTHON_OPTION $MKL numpy"
     if [[ -z $(pip list -o --pre | grep numpy | \
             grep -E "[0-9]rc[0-9]|[0-9][ab][0-9]") ]]; then
         # We want to stop the script if there isn't a pre-release available,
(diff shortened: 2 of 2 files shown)


travis/setup_dependencies_common.sh
```diff
@@ -341,13 +344,13 @@ if [[ $NUMPY_VERSION == dev* ]]; then
     # We then install Numpy itself at the bottom of this script
     export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $MKL"
 elif [[ $NUMPY_VERSION == stable ]]; then
-    retry_on_known_error conda install $QUIET --no-pin numpy=$LATEST_NUMPY_STABLE $MKL
     export NUMPY_OPTION="numpy=$LATEST_NUMPY_STABLE"
-    export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION numpy=$LATEST_NUMPY_STABLE $MKL"
+    export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $NUMPY_OPTION $MKL"
+    NUMPY_INSTALL="conda install $QUIET --no-pin $PYTHON_OPTION $NUMPY_OPTION $MKL"
 elif [[ $NUMPY_VERSION == pre* ]]; then
-    retry_on_known_error conda install $QUIET --no-pin $MKL numpy
```
bsipocz: here it doesn't matter, for prereleases numpy will be reinstalled anyway at the end.

travis/setup_dependencies_common.sh
```diff
@@ -357,14 +360,37 @@ elif [[ $NUMPY_VERSION == pre* ]]; then
...
     export CONDA_INSTALL="conda install $QUIET $PYTHON_OPTION $MKL"
 fi
 
+# try to install numpy:
+if [[ ! -z $NUMPY_INSTALL ]]; then
+    retry_on_known_error $NUMPY_INSTALL || { \
+        if [[ -z $NUMPY_OPTION ]]; then
+            PIP_NUMPY_OPTION="numpy"
+        else
+            if [[ $NUMPY_OPTION =~ ^numpy$is_eq_float$ ]]; then
```
bsipocz: so you forbid here to install 1.15.2?

bsipocz:
Hmm. Arguably I'm not up to date with the latest and greatest conda, so my concern may not be relevant any more. But, I fear, that conda still doesn't play together nicely with pip. So if we pip install numpy, and there are other dependencies that rely on numpy, conda would mess up the installs.

An easy hack (but an ugly one) could be to keep "stable" 1.15 for python3.5 (we have a similar hack for astropy versions).

travis/setup_dependencies_common.sh
zemanj: Sure, this will never fail. But there's no reason for not having the numpy installation in a single place either. IMHO this is cleaner.

travis/setup_dependencies_common.sh
zemanj: You're right, I forgot to handle cases with exact version requirements further below.
This particular check is to add a a wildcard for pip so that it behaves like conda, i.e., to transform the conda requirement `numpy=1.16` into the corresponding pip requirement `numpy==1.16.*`.
I'll add an additional check for arbitrary versions.

bsipocz: we already have some workarounds for those when falling back to pip install for conda dependency packages. Maybe simply adding numpy to the CONDA_DEPENDENCIES list would work (and probably should have done a long time ago, but this whole pile of mess just evolved very organically hacking the hacks of an early hack).

bsipocz: (numpy originally was handled separately as that was the only dependency astropy had. then we added layers and layers of stuff on top while should have started from the beginning again. But then, now, given that we can't really see what happens to travis in the near future probably it's not worth spending too much time with big refactoring.)

bsipocz:
None of the travis tests is actually testing this change. Could you maybe add "NUMPY_VERSION=stable PYTHON_VERSION=3.5" to build numpy 29 to see that this works as intended?

```
        - os: linux
          env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
               CONDA_DEPENDENCIES_FLAGS='--no-deps'
```

travis/setup_dependencies_common.sh
zemanj: I'm (unfortunately) very familiar with pip fallbacks... :wink: 
Using the same functionality here is not an option, though, since then, one would have to disallow numpy installs via conda, leading to a massive amount of pip fallbacks for the rest of the `CONDA_DEPENDENCIES`.
Adding numpy to `CONDA_DEPENDENCIES` will cause single conda installs for all packages as well because the conda install with all packages at once *will* fail. One-by-one installs are a (very) sub-optimal solution since package (inter-)dependencies are then resolved incrementally, which is dead slow...
So, I think that once we know that numpy can't be installed via conda, there is no reason to not add numpy to `PIP_DEPENDENCIES`.

travis/setup_dependencies_common.sh
```diff
@@ -357,14 +360,43 @@ elif [[ $NUMPY_VERSION == pre* ]]; then
...
+if [[ ! -z $NUMPY_INSTALL ]]; then
+    retry_on_known_error $NUMPY_INSTALL || { \
+        if [[ -z $NUMPY_OPTION ]]; then
+            PIP_NUMPY_OPTION="numpy"
+        else
+            # add wildcard for float-like version specs:
+            if [[ $NUMPY_OPTION =~ ^numpy$is_eq_float$ ]]; then
+                PIP_NUMPY_OPTION="numpy==${NUMPY_OPTION#*=}.*"
+            # use exact version definitions as is:
+            elif [[ $NUMPY_OPTION =~ ^numpy=.* ]]; then
```
zemanj: This should do the trick

.travis.yml
```diff
@@ -190,6 +190,9 @@ matrix:
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
                CONDA_DEPENDENCIES_FLAGS='--no-deps'
+               NUMPY_VERSION=stable PYTHON_VERSION=3.5
+               CONDA_CHANNELS='conda-forge'
```
bsipocz: why conda-forge?

zemanj: because conda-forge misses the numpy 1.16 build for py35

bsipocz: but the exercise here is to check whether your patch works, isn't it?

zemanj: Right?! This is to test if the "pip fallback" works as intended for a missing numpy build. The other Travis tests also work as intended, so my PR didn't brake anything if a suitable build is available.

bsipocz: Oh, sorry, I misread your comment. The default channel doesn't have 1.16 for python 3.5 either, and that preoccupied my thoughts.

zemanj: * In test 4, for example, numpy is [successfully installed via conda](https://travis-ci.org/astropy/ci-helpers/jobs/502868436#L1644) and the fallback is not triggered. So the fix works as intended.
* In test 29, [installing numpy via conda fails](https://travis-ci.org/astropy/ci-helpers/jobs/502868461#L660), [numpy is added to `PIP_DEPENDENCIES`](https://travis-ci.org/astropy/ci-helpers/jobs/502868461#L667), and is later [installed via pip](https://travis-ci.org/astropy/ci-helpers/jobs/502868461#L706).
* Do we need an additional test for an exact version specification (1.16.2)?

zemanj: >The default channel doesn't have 1.16 for python 3.5 either

Sorry I wasn't sure about this, so I used conda-forge.

zemanj: Hmm... I think the test should explicitly state the numpy version instead of using "stable" since that's subject to change... I'm pretty sure that Python 3.5 support for numpy will be dropped in ~2020 completely, which means that the pip install will also fail for a then "latest stable" numpy version.

zemanj: I pinned the version in test 29 to 1.16 and added an additional test for 1.16.1

bsipocz: > I'm pretty sure that Python 3.5 support for numpy will be dropped in ~2020 completely

you're very optimistic that we still have travis in 2020 :)

travis/setup_dependencies_common.sh
bsipocz: at some point I think we need to also check whether there is anything in CONDA_DEPENDENCIES, and if not, but there is PIP_DEPENDENCIES, then just install numpy from pip (we already have a few packages where we switched to pip only installs as it's way faster). Anyway, it's into the weeds for this PR, but maybe something for the future.

bsipocz:
Thanks @zemanj, I'm so happy that other projects are also finding ci-helpers useful to the point to contribute back!

zemanj:
My pleasure. Thank you for your quick response, I appreciate it!

Issue #353, opened 2018-11-29
Consider updating miniconda version


bsipocz:
4.5.11 now comes with python3.7 and as we use that as default more and more places it would make sense to avoid one more python install.

djhoese:
I was just about to make an issue for this, but conda 4.6 is available (not through miniconda installer yet). More importantly 4.6.3 fixes an issue where package "activation" scripts are now run after the package is installed. Without this packages like pyproj don't have their environment variables set correctly unless you deactivate and reactivate the test environment.

Would it be possible to include a `conda update conda` command in `ci-helpers`? See https://github.com/conda-forge/pyproj-feedstock/issues/29 for details.

bsipocz:
You can have any conda version number you like, just use `CONDA_VERSION` in the travis config.

djhoese:
Is that documented somewhere? I don't see it in the README?

djhoese:
Also is there a way to say "the latest version"?

bsipocz:
Hmm, there are missing pieces here and there. If you have time to open a PR, it would be more than welcome for both the README and the latest version. 

E.g. changing the logic to something like:
```
if [[ -z $CONDA_VERSION ]]; then
    CONDA_VERSION=4.5.12
elif [[ $CONDA_VERSION == latest ]] || [[ $CONDA_VERSION == stable ]]; then
    CONDA_VERSION=
fi
```

PR #389, opened 2019-03-13
Add CONDA_VERSION='stable' option

djhoese:
As mentioned in https://github.com/astropy/ci-helpers/issues/353 this adds the option of specifying the CONDA_VERSION as ``stable`` to get the newest version.

Honestly, I don't think this will work for appveyor the way I have it set up, but wanted to get this started first.

@bsipocz you suggested including "latest" as an option but after looking at the other available environment variables it seemed like "stable" was semi-consistent. If you want me to add "latest" as an option too let me know (and tell me how to do powershell OR conditions).


bsipocz:
Thank you! I suppose we could test this by just modifying one of the existing travis jobs.

djhoese:
Right now the scripts for travis write to a pinning file (I think) so for the `stable` case you get `conda ` (note the extra space). For appveyor it runs `conda install conda=` which I *think* will fail. Is there a way to set something on the right hand side to mean "stable" or "latest"? Or does the code need to be modified to have another if statement (if conda version specify it, else update conda)?

bsipocz:
Ideally the appveyor one could be changed to follow the travis logic, namely to put the versions into the pin file. It seems thought that besides the python version nothing ends up there.

Otherwise I think it indeed needs another if, or moving the `CONDA_VERSION` check to around line 263 and add do the conda installs within that single if statement.

I don't really have preferences here, in our projects we're moving away from appveyor in the medium term given that both travis and azure provides windows VMs so I try to minimize the time spent with these horrid powershell scripts we have.

djhoese:
@bsipocz Are you against the latest version of conda being the default instead of 4.5.10 as it is now?

bsipocz:
> @bsipocz Are you against the latest version of conda being the default instead of 4.5.10 as it is now?

I'm afraid yes. We started with that setup and it caused terrible headaches as issues came up when we didn't really have time to fix them yet it was critical to fix so needed to drop everything else. 
The current policy is to update that version number from time whenever we know we have some extra bandwidth to deal with possible issues (and can test the new default in a handful of packages prior, typically using astropy and e.g. photutils, astroquery/gammapy and asking people who manage multiple packages to also have a look, e.g. https://github.com/astropy/ci-helpers/pull/273).

djhoese:
Understood. Also I think my last commit may not be needed. I think `conda install conda=` will install the most recent version. I'll try to look at it this weekend or Monday.

djhoese:
@bsipocz I was trying to look through the travis tests (I should probably modify/add an appveyor env too) and I wasn't really sure what one to add a `CONDA_VERSION=stable` too for fear of changing the purpose/validity of the test. Ideas?

djhoese:
I was just fixing your `os.environ('CONDA_VERSION')` but then you force pushed. Let me know if you want me to fix anything, otherwise I'll wait for your reply.

bsipocz:
Yes, sorry, haven't noticed that. There are still a few issues, I'm trying to fix them.

djhoese:
I completely forgot about this. @bsipocz is there anything I can do to help out with this?

astrofrog:
@djhoese - could you try rebasing this as a fair bit has changed, including dropped support for AppVeyor?

djhoese:
Not a lot of time but maybe I'll try later today or later this week.

djhoese:
Done. Accidentally forgot to fetch the latest upstream master during my first merge. Everything is up to date now and I just ignored any appveyor changes I made.

djhoese:
@astrofrog Any suggestions for the failing tests? At least two of them don't seem like I caused them. The other one is CONDA_VERSION related (LooseVersion has no version attribute), but the if statement above it specifically doesn't run this code for `'stabie'` CONDA_VERSION *and* the if statement above that isn't even supposed to run for Python 3.7.

pllim:
Astropy Project has decided to move away from Travis CI, therefore we regretfully are unable to accept this PR any longer. We really appreciate your contributions and hope you would continue to contribute to the project in other ways. Thank you and sorry for any inconvenience caused.

Please see https://groups.google.com/g/astropy-dev/c/45frnTEAX-U for more information.

PR #396, opened 2019-06-15
Remove astropy-ci-extras

bsipocz:
It hasn't been used for a while and has been archived. We don't really support versions any more that would rely on it.


bsipocz:
(appveyor is not affected in this PR, cancelling runs)

PR #403, opened 2019-07-31
Activate conda on Windows the official™️ way

mwcraig:
Builds are failing on appveyor very early in the setup process. The issue is https://github.com/conda/conda/issues/8836, though the most useful bits are at https://github.com/conda/conda/issues/8836#issuecomment-508815777 and the changes are based on https://github.com/geopandas/geopandas/pull/1029/files

I don't have a windows machine handy, so we'll see what appveyor thinks about this...

PR #405, opened 2019-08-02
Switch to conda 4.7 as the default conda  and use flexible in place of true

mwcraig:
We should consider using strict channel priority for all the builds because it can speed up installation considerably. For now I've just replace `true` with `strict`.

I am quite sure some of the appveyor builds are going to break because they use quite old python versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,7 +163,7 @@ matrix:
         # Limit attrs version as newer ones break old pytest
         - os: linux
           env: SETUP_CMD='build_docs' PYTHON_VERSION=2.7 ASTROPY_VERSION='stable'
-               CONDA_DEPENDENCY='attrs<19.2'
+               CONDA_DEPENDENCIES='attrs<19.2'
 
         # JOB .15
         - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,11 +60,6 @@ matrix:
         #    is 10.0.1 while the latest is 18.0 at the time of writing this test.
         # -> The latest version of pip as of 2019-08-15 on default channels is 19.2.2
         #    for Python >=3.6, so bump the pip version check to minimum of 19.
-        # Updated these versions because of availability of packages on the default
-        # channels:
-        #
-        #    PYTHON_VERSION 3.5 → 3.6
-        #    pip minimum version >17 → >18
         - os: linux
           env: SETUP_CMD='build_docs' SPHINX_VERSION='>1.6'
                PYTHON_VERSION=3.6 DEBUG=True
@@ -127,26 +122,12 @@ matrix:
 
         # JOB .11
         # -> Setting pytest version should be recognized in the next two jobs
-        #
-        # Changed version numbers in 2019 because older versions are no longer
-        # available in default conda channels:
-        #
-        #    matplotlib 1.5.0 → 2.2.0
-        #    pytest 2.7.3 → 4.6.4
         - os: linux
           env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib>2.0,<=2.2.0'
                DEBUG=True PYTHON_VERSION=2.7 PYTEST_VERSION=4.6.4
                TEST_CMD='conda list; python -c "import matplotlib; assert matplotlib.__version__==\"2.2.0\"; import pytest; assert pytest.__version__.startswith(\"4.6.4\")"'
 
         # JOB .12
-        # Changed version numbers in 2019 because older versions are no longer
-        # available in default conda channels:
-        #
-        #    PYTHON_VERSION 3.5 → 3.6
-        #    MATPLOTLIB_VERSION 1.4 → 3.0 (and expected installed version 3.0.3)
-        #    SPHINX_VERSION 1.4.1 → 1.8.4
-        #    PYTEST_VERSION 3.0 → 5.0
-        #
         - os: linux
           env: SETUP_CMD='build_sphinx' SPHINX_VERSION=1.8.4 DEBUG=True
                MATPLOTLIB_VERSION=3.0 DEBUG=True PYTHON_VERSION=3.6 PYTEST_VERSION=5.0
@@ -212,8 +193,8 @@ matrix:
         #    can make sure the conda config was correctly updated.
         # -> Have a package dependency with alphanumerical name
         #
-        # -> The old cahnnel priority "true" is equivalent to the new priority
-        #    strict.
+        # -> The old channel priority "true" is equivalent to the new priority
+        #    "flexible".
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=flexible
                CONDA_DEPENDENCIES='jinja2=2.8 h5py matplotlib' DEBUG=True
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
           addons:
               apt:
                   packages: gfortran
-          env: NUMPY_VERSION=stable CONDA_DEPENDENCIES='healpy matplotlib scikit-learn scipy Pillow'
+          env: NUMPY_VERSION=stable CONDA_DEPENDENCIES='healpy matplotlib scikit-learn scipy pillow'
                CONDA_CHANNELS='conda-forge' DEBUG=True
                SUNPY_VERSION=dev MATPLOTLIB_VERSION='dev'
                SCIKIT_LEARN_VERSION='dev'
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ matrix:
                PYTEST_VERSION="<3.6.2"
 
         # JOB .7
-        # -> Have non Python package (libcxx) in dependency list when
+        # -> Have non Python package (libtiff) in dependency list when
         #    falling back on pip installing the dependencies (as ipyevents is
         #    missing from listed channels) to test the install one-by-one.
         # -> Have package dependency that matches partially another package
@@ -109,8 +109,8 @@ matrix:
           env: SETUP_CMD='test --coverage' PYTHON_VERSION=3.6
                TEST_CMD='conda list; py.test test_env.py && coverage run test_env.py; python -c "import ipyevents; assert ipyevents.__version__.startswith(\"0.4\")"'
                CONDA_CHANNELS=''
-               CONDA_DEPENDENCIES='libcxx pytest pytest-cov ipyevents'
-               DEBUG=True NUMPY_VERSION=1.15 ASTROPY_VERSION=stable
+               CONDA_DEPENDENCIES='libtiff pytest pytest-cov ipyevents'
+               DEBUG=True ASTROPY_VERSION=stable
                IPYEVENTS_VERSION='<0.5,>=0.4'
 
         # JOB .8
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,16 +13,20 @@ env:
         - SETUP_CMD='test'
         - TEST_CMD='py.test test_env.py'
         - PYTHON_VERSION=3.6
+        - DEBUG=True
 
     matrix:
         - SETUP_CMD='egg_info' TEST_CMD='python --version'
 
 matrix:
     include:
+        # JOB .1
         # pytest pip fallback, conda package for 3.1 is not available for py3.4
         - os: linux
           env: PYTHON_VERSION=3.4 PYTEST_VERSION=3.1
 
+        # JOB .33 (because it is under allowed failures)
+        # NOTE: IT DOES NOT CURRENTLY FAIL
         # Try pytest pip fallback as conda package for 3.1 is not available
         # for py3.4 but then fail as PIP_FALLBACK is False. Listing this
         # under the allowed failures as there is no way at the moment to
@@ -30,9 +34,11 @@ matrix:
         - os: linux
           env: PYTHON_VERSION=3.4 PYTEST_VERSION=3.1 PIP_FALLBACK=false
 
+        # JOB .2
         - os: windows
           env: PYTHON_VERSION=3.7
 
+        # JOB .3
         # -> tests sunpy dev
         # -> Test falling back on pip when conda install is not working.
         #    Here healpy is not available with numpy >=1.13
@@ -48,14 +54,23 @@ matrix:
                SUNPY_VERSION=dev MATPLOTLIB_VERSION='dev'
                SCIKIT_LEARN_VERSION='dev'
 
+        # JOB .4
         # -> Testing whether pip install fallback is working for sphinx/matplotlib in the docs builds
         # -> Test we have a recent pip version, the newest on the defaults channel
         #    is 10.0.1 while the latest is 18.0 at the time of writing this test.
+        # -> The latest version of pip as of 2019-08-15 on default channels is 19.2.2
+        #    for Python >=3.6, so bump the pip version check to minimum of 19.
+        # Updated these versions because of availability of packages on the default
+        # channels:
+        #
+        #    PYTHON_VERSION 3.5 → 3.6
+        #    pip minimum version >17 → >18
         - os: linux
           env: SETUP_CMD='build_docs' SPHINX_VERSION='>1.6'
-               PYTHON_VERSION=3.5 DEBUG=True
-               TEST_CMD='python -c "import sphinx; from distutils.version import LooseVersion; assert LooseVersion(sphinx.__version__)>LooseVersion(\"1.5\"); import pip; assert int(pip.__version__.split(\".\")[0])>17"'
+               PYTHON_VERSION=3.6 DEBUG=True
+               TEST_CMD='python -c "import sphinx; from distutils.version import LooseVersion; assert LooseVersion(sphinx.__version__)>LooseVersion(\"1.5\"); import pip; assert int(pip.__version__.split(\".\")[0])>18"'
 
+        # JOB .5
         # -> tests sunpy stable
         # -> Starting with the dev versions as they take the longest to run. We
         #    deliberately test with Numpy 1.16 to check that installing Astropy
@@ -64,6 +79,7 @@ matrix:
           env: SETUP_CMD='test' NUMPY_VERSION=1.16 ASTROPY_VERSION=dev
                PYTHON_VERSION=3.6 SUNPY_VERSION=stable CONDA_CHANNELS='conda-forge'
 
+        # JOB .6
         # -> For the Numpy dev build, we also specify a conda package that
         #    depends on Numpy to make sure that Numpy dev is used (because
         #    conda will also install Numpy stable). It's also important to
@@ -76,139 +92,189 @@ matrix:
                CONDA_DEPENDENCIES='scipy'
                PYTEST_VERSION="<3.6.2"
 
+        # JOB .7
         # -> Have non Python package (openjpeg) in dependency list when
-        #    falling back on pip installing the dependencies (as photutils is
-        #    missing from listed channels) to test the install one-by-one.
+        #    falling back on pip installing the dependencies (as pyqt5 is
+        #    missing from listed channels where it is called pyqt) to test the install one-by-one.
         # -> Have package dependency that matches partially another package
         #    (pytest vs pytest-cov) when falling back to pip install
         # -> testing whether version numbers are respected by pip install fallback
+        # The following versions were updated in 2019 because the old versions
+        # were leading to issues because anaconda deprecated the old free channel:
+        #
+        #    PYTHON_VERSION 3.4 → 3.6
+        #    NUMPY_VERSION 1.10 → 1.16
+        #
         - os: linux
-          env: SETUP_CMD='test --coverage' PYTHON_VERSION=3.4
-               TEST_CMD='py.test test_env.py && coverage run test_env.py; python -c "import sklearn; assert sklearn.__version__.startswith(\"0.19\")"'
+          env: SETUP_CMD='test --coverage' PYTHON_VERSION=3.6
+               TEST_CMD='conda list; py.test test_env.py && coverage run test_env.py; python -c "import sklearn; assert sklearn.__version__.startswith(\"0.19\")"'
                CONDA_CHANNELS='astrofrog conda-forge'
                CONDA_DEPENDENCIES='scipy pyqt5 openjpeg photutils pytest pytest-cov scikit-learn'
-               DEBUG=True NUMPY_VERSION=1.10 ASTROPY_VERSION=stable
+               DEBUG=True NUMPY_VERSION=1.15 ASTROPY_VERSION=stable
                SCIKIT_LEARN_VERSION='<0.20,>=0.19'
 
+        # JOB .8
         # -> The pre builds should run the testing phase only when a pre
         #    release is available on pypi, otherwise it should pass
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=pre
+        # JOB .9
         - os: linux
           env: SETUP_CMD='test' ASTROPY_VERSION=pre DEBUG=True
+        # JOB .10
         - os: linux
           env: SUNPY_VERSION=prerelease CONDA_CHANNELS='conda-forge'
 
+        # JOB .11
         # -> Setting pytest version should be recognized in the next two jobs
+        #
+        # Changed version numbers in 2019 because older versions are no longer
+        # available in default conda channels:
+        #
+        #    matplotlib 1.5.0 → 2.2.0
+        #    pytest 2.7.3 → 4.6.4
         - os: linux
-          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib<=1.5.0'
-               DEBUG=True PYTHON_VERSION=2.7 PYTEST_VERSION=2.7.3
-               TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__==\"1.5.0\"; import pytest; assert pytest.__version__.startswith(\"2.7.3\")"'
+          env: SETUP_CMD='build_sphinx' CONDA_DEPENDENCIES='matplotlib>2.0,<=2.2.0'
+               DEBUG=True PYTHON_VERSION=2.7 PYTEST_VERSION=4.6.4
+               TEST_CMD='conda list; python -c "import matplotlib; assert matplotlib.__version__==\"2.2.0\"; import pytest; assert pytest.__version__.startswith(\"4.6.4\")"'
 
+        # JOB .12
+        # Changed version numbers in 2019 because older versions are no longer
+        # available in default conda channels:
+        #
+        #    PYTHON_VERSION 3.5 → 3.6
+        #    MATPLOTLIB_VERSION 1.4 → 3.0 (and expected installed version 3.0.3)
+        #    SPHINX_VERSION 1.4.1 → 1.8.4
+        #    PYTEST_VERSION 3.0 → 5.0
+        #
         - os: linux
-          env: SETUP_CMD='build_sphinx' SPHINX_VERSION=1.4.1 DEBUG=True
-               MATPLOTLIB_VERSION=1.4 DEBUG=True PYTHON_VERSION=3.5 PYTEST_VERSION=3.0
-               TEST_CMD='python -c "import sphinx; assert sphinx.__version__==\"1.4.1\"; import pytest; pytest.__version__.startswith(\"3.0\"); import matplotlib; assert matplotlib.__version__=="\""1.4.3"\"'
+          env: SETUP_CMD='build_sphinx' SPHINX_VERSION=1.8.4 DEBUG=True
+               MATPLOTLIB_VERSION=3.0 DEBUG=True PYTHON_VERSION=3.6 PYTEST_VERSION=5.0
+               TEST_CMD='python -c "import sphinx; assert sphinx.__version__==\"1.8.4\"; import pytest; pytest.__version__.startswith(\"5.0\"); import matplotlib; assert matplotlib.__version__=="\""3.0.3"\"'
 
+        # JOB .13
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib=1.5.0 sphinx'
                MATPLOTLIB_VERSION=1.5.1 DEBUG=True PYTHON_VERSION=3.5
                TEST_CMD='python -c "import matplotlib; assert matplotlib.__version__=="\""1.5.1"\"'
 
+        # JOB .14
         # -> astropy stable should stay on the 2.0.x branch
         # Limit attrs version as newer ones break old pytest
         - os: linux
           env: SETUP_CMD='build_docs' PYTHON_VERSION=2.7 ASTROPY_VERSION='stable'
                CONDA_DEPENDENCY='attrs<19.2'
 
+        # JOB .15
         - os: linux
           env: MAIN_CMD='pep8'
                TEST_CMD='pep8 --version'
 
+        # JOB .16
         - os: linux
           env: MAIN_CMD='pycodestyle'
                TEST_CMD='pycodestyle --version'
 
+        # JOB .17
         - os: linux
           env: MAIN_CMD='flake8'
                TEST_CMD='flake8 --version'
 
+        # JOB .18
         - os: linux
           env: MAIN_CMD='pylint'
                TEST_CMD='pylint --version'
 
+        # JOB .19
         - os: linux
           env: SETUP_CMD="--cov package"
                TEST_CMD='python -c "import pytest_cov"'
 
+        # JOB .20
         # -> Don't test osx with LTS as we currently build LTS only for linux.
         - os: osx
           env: SETUP_CMD='test' ASTROPY_VERSION=stable
                CONDA_DEPENDENCIES='requests'
 
-        # -> Note that setting CONDA_CHANNEL_PRIORITY to False here is unrelated
+        # JOB .21
+        # -> Note that setting CONDA_CHANNEL_PRIORITY to disabled here is unrelated
         #    to the rest of the build and won't change anything, except that we
         #    can make sure the conda config was correctly updated.
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=1.10 DEBUG=True
                SCIKIT_IMAGE_VERSION=0.11 PYTHON_VERSION=3.5
                CONDA_DEPENDENCIES='Cython numpy scikit-image'
-               CONDA_CHANNEL_PRIORITY=False
+               CONDA_CHANNEL_PRIORITY=disabled
 
-        # -> Note that setting CONDA_CHANNEL_PRIORITY to strict here is unrelated
+        # JOB .22
+        # -> Note that setting CONDA_CHANNEL_PRIORITY to flexible here is unrelated
         #    to the rest of the build and won't change anything, except that we
         #    can make sure the conda config was correctly updated.
         # -> Have a package dependency with alphanumerical name
+        #
+        # -> The old cahnnel priority "true" is equivalent to the new priority
+        #    strict.
         - os: linux
-          env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=strict
+          env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=flexible
                CONDA_DEPENDENCIES='jinja2=2.8 h5py matplotlib' DEBUG=True
                ASTROPY_VERSION=1.0 CONDA_CHANNELS='astropy'
                PYTHON_VERSION='3.5'
 
+        # JOB .23
         # -> Have empty string as dependency list when falling back on pip
         # Limit attrs version as newer ones break old pytest
         - os: linux
           env: SETUP_CMD='test' NUMPY_VERSION=stable ASTROPY_VERSION=lts
                CONDA_DEPENDENCIES='attrs<19.2'
 
+        # JOB .24
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib=1.5'
                TEST_CMD='python -c "import matplotlib; assert int(matplotlib.__version__[-1])>0"'
 
+        # JOB .25
         - os: linux
           env: SETUP_CMD='test --parallel'
                TEST_CMD='py.test -n 4 test_env.py'
 
+        # JOB .26
         - os: linux
           env: SETUP_CMD='test --numprocesses'
                TEST_CMD='py.test --numprocesses 4 test_env.py'
 
+        # JOB .27
         - os: linux
           env: SETUP_CMD='test --open-files'
 
+        # JOB .28
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
                CONDA_DEPENDENCIES_FLAGS='--no-deps'
                NUMPY_VERSION=1.16 PYTHON_VERSION=3.5
 
+        # JOB .29
         - os: linux
           env: SETUP_CMD='test' CONDA_DEPENDENCIES='matplotlib'
                CONDA_DEPENDENCIES_FLAGS='--no-deps'
                NUMPY_VERSION=1.16.1 PYTHON_VERSION=3.5
 
+        # JOB .30
         - os: linux
           env: SETUP_CMD='test' PIP_DEPENDENCIES='pyparsing requests'
                PIP_DEPENDENCIES_FLAGS='--log pip_install.log'
 
+        # JOB .31
         - os: linux
           env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml'
                TEST_CMD="python -c $'import matplotlib\ntry:\n    import scipy\nexcept ImportError:\n    pass\nelse:\n    raise RuntimeError'"
 
+        # JOB .32
         - os: linux
           env: SETUP_CMD='test' CONDA_ENVIRONMENT='conda_environment.yml' CONDA_DEPENDENCIES="scipy"
                TEST_CMD="python -c $'import matplotlib\nimport scipy'"
 
     allow_failures:
+        # JOB .33
         - os: linux
           env: PYTHON_VERSION=3.4 PYTEST_VERSION=3.1 PIP_FALLBACK=false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -155,12 +155,12 @@ matrix:
                CONDA_DEPENDENCIES='Cython numpy scikit-image'
                CONDA_CHANNEL_PRIORITY=False
 
-        # -> Note that setting CONDA_CHANNEL_PRIORITY to True here is unrelated
+        # -> Note that setting CONDA_CHANNEL_PRIORITY to strict here is unrelated
         #    to the rest of the build and won't change anything, except that we
         #    can make sure the conda config was correctly updated.
         # -> Have a package dependency with alphanumerical name
         - os: linux
-          env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=True
+          env: SETUP_CMD='test' NUMPY_VERSION=1.10 CONDA_CHANNEL_PRIORITY=strict
                CONDA_DEPENDENCIES='jinja2=2.8 h5py matplotlib' DEBUG=True
                ASTROPY_VERSION=1.0 CONDA_CHANNELS='astropy'
                PYTHON_VERSION='3.5'
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,25 +93,25 @@ matrix:
                PYTEST_VERSION="<3.6.2"
 
         # JOB .7
-        # -> Have non Python package (openjpeg) in dependency list when
-        #    falling back on pip installing the dependencies (as pyqt5 is
-        #    missing from listed channels where it is called pyqt) to test the install one-by-one.
+        # -> Have non Python package (libcxx) in dependency list when
+        #    falling back on pip installing the dependencies (as ipyevents is
+        #    missing from listed channels) to test the install one-by-one.
         # -> Have package dependency that matches partially another package
         #    (pytest vs pytest-cov) when falling back to pip install
         # -> testing whether version numbers are respected by pip install fallback
-        # The following versions were updated in 2019 because the old versions
-        # were leading to issues because anaconda deprecated the old free channel:
         #
-        #    PYTHON_VERSION 3.4 → 3.6
-        #    NUMPY_VERSION 1.10 → 1.16
+        # The list of dependencies was considerably simplified to speed up the
+        # (expected) conda failure.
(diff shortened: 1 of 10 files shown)


mwcraig:
Set a new personal best in number of tests that failed 🤣 😭

bsipocz:
@mwcraig - updating the conda version pin has always been good fun :)

.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
         - SETUP_CMD='test'
         - TEST_CMD='py.test test_env.py'
         - PYTHON_VERSION=3.6
+        - MINICONDA_VERSION=4.7.10
```
bsipocz: could you rather change the default in the `travis/setup_conda_*` files?

mwcraig:
Remaining issues:

+ [ ] Some jobs are failing because the name for no channel priority is now `disabled` instead of `false`.
+ [ ] `conda` is not being activated properly on windows.
+[ ] Several tests fail because the package `_license` is not available (all of the pythons < 3.7). Solution is to add the free channel back in those cases, see https://github.com/conda/conda/issues/8863#issuecomment-509746792 and https://www.anaconda.com/why-we-removed-the-free-channel-in-conda-4-7/

Working on these changes now...

mwcraig:
A question:

+ Why is `--no-channel-priority` used at https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L214?

mwcraig:
This is currently stalled on https://github.com/conda/conda/issues/9052

In a nutshell, the pinning in the file are being ignored. That can be fixed by figuring out which packages/versions are so old that they are only in the `free` channel and adding the `free` channel back in those cases.

I'd rather not add the `free` channel in all of the builds because it slows the solver down a lot.

mwcraig:
One more question:

Why is the latest stable numpy pinned at 1.16 at https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L109?

bsipocz:
> Why is the latest stable numpy pinned at 1.16 

as numpy 1.17 is still not available on the default channels

bsipocz:
> A question:
> 
>     * Why is `--no-channel-priority` used at https://github.com/astropy/ci-helpers/blob/master/travis/setup_dependencies_common.sh#L214?

I suppose that is obsolete now that we don't use versions from `astropy-ci-extras`:
https://github.com/bsipocz/ci-helpers/commit/50fbfb9a105383cd507c3b8f42db3cdf9b6bff14

mwcraig:
@bsipocz -- any idea why [this fails](https://github.com/astropy/ci-helpers/pull/405/files#diff-ee1b658811190e8695c7b64c9d889e3aR329) on travis [like on this line](https://travis-ci.org/astropy/ci-helpers/jobs/569093573#L1549)?

travis/setup_dependencies_common.sh
```diff
@@ -313,7 +321,55 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
     fi
 fi
 
+if [[ ! -z $CONDA_DEPENDENCIES ]]; then
+    # Do a dry run of the conda install here to make sure that pins are
+    # ACTUALLY being respected. This will become unnecessary when
+    # https://github.com/conda/conda/issues/9052
+    # is fixed
+    dry_run_bad=$(conda install --dry-run $CONDA_DEPENDENCIES 2>&1 | grep -c "conflicts with explicit specs")
```
bsipocz:
```suggestion
    dry_run_bad=`conda install --dry-run $CONDA_DEPENDENCIES 2>&1 | grep -c "conflicts with explicit specs"`
```

bsipocz:
@mwcraig - It executes 0 rather than assignes it as a numerical value to `dry_run_bad`

travis/setup_dependencies_common.sh
```diff
@@ -313,7 +321,55 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
...
+    echo "dry_run_bad is $dry_run_bad"
+    if [[ $dry_run_bad -gt 0 ]]; then
```
bsipocz: rather than having `dry_run_bad` defined, the usual logic all over this file is to compare it with test -z:

``if [[ ! -z $(conda install --dry-run $CONDA_DEPENDENCIES 2>&1 | grep  "conflicts with explicit specs") ]]; then 

not sure whether it's clearer or not though, and I have to admit this file is my worst pile of hacks ever that just keep growing organically 🙈

travis/setup_dependencies_common.sh
```diff
@@ -313,7 +321,55 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
...
+        # channel priority
+        if [[ ! -z $CONDA_CHANNEL_PRIORITY && $CONDA_CHANNEL_PRIORITY == strict ]]; then
+            # If the channel priority is strict we should fail instead of silently
+            # changing how the solve is done.
+            echo "cannot solve this environment with pinnings and strict channel priority"
+            exit 1
+        fi
+
+        # Try the dry run again, fail if pinnings are still ignored
+        dry_run_bad=$(conda install --dry-run $CONDA_DEPENDENCIES 2>&1 | grep -c "conflicts with explicit specs")
```
bsipocz: you need the backticks here, too
```suggestion
        dry_run_bad=`conda install --dry-run $CONDA_DEPENDENCIES 2>&1 | grep -c "conflicts with explicit specs"`
```

djhoese:
Anything I can do to help with this migration? Will this improve appveyor tests or is #403 still going to be a difficult task?

bsipocz:
> Anything I can do to help with this migration? 

I have to admit I lost track where we stand with this.

mwcraig:
This won't help with appveyor, though it might make Windows work for you. 

### appveyor 

There are a number of problems with the current appveyor setup, unfortunately, the trickiest of which is that the PowerShell script that is supposed to rerun if the failure is a conda HTTP error isn't working and I know nothing about PowerShell. 

The more straightforward issue is that newer versions of conda require a different activation mechanism on Windows than they have in the past.

### this pr

In addition, there is a bug in the current conda release which results in pinnings in a file being overridden even when there is no pin on the command line (see https://github.com/astropy/ci-helpers/pull/405#issuecomment-5182397640).

I was trying to catch those errors and fix them by adding the necessary channels at https://github.com/astropy/ci-helpers/pull/405/files#diff-ee1b658811190e8695c7b64c9d889e3aR324

That fails [like on this line](https://travis-ci.org/astropy/ci-helpers/jobs/569093573#L1549) even though it works locally.

### now what 🤷‍♂️

1. Always add the free channel back for now. That may fix some of the pin-breaking we are seeing.
2. Add the free channel back for older versions of python; that way newer versions solve faster. Skip the new, fancy, broken logic.
3. Wait for the conda bug to be fixed. No ETA on that.
4. Other...?

### pushing a quick thing for debugging...

If there is a failure in the `conda --dry-run` I'm trying we won't see the message because I'm redirecting standard out. I'll try running it on a line by itself to see what the output is.

djhoese:
Fixing that shell exit fixed my runs for Linux and Mac: https://travis-ci.org/pytroll/pyresample/builds/571326372

travis/setup_conda_windows.sh
```diff
@@ -6,12 +6,14 @@ if [[ $DEBUG == True ]]; then
...
 
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=4.5.4
+    MINICONDA_VERSION=4.6.14
 fi
 
 echo "installing miniconda3"
 choco install miniconda3 --params="'/AddToPath:1'" --version="$MINICONDA_VERSION";
-export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/:$PATH";
+export PATH="/c/tools/miniconda/bin:$PATH"
```
djhoese: I think this needs to be `miniconda3`

djhoese:
```suggestion
export PATH="/c/tools/miniconda3/bin:$PATH"
```

mwcraig: Oh yeah oops 😮

djhoese: I'm testing it in my own fork. Didn't seem to work the first time, so I added DEBUG and am still waiting.

djhoese: Does this need to be `/mnt/c`? https://superuser.com/questions/1113385/convert-windows-path-for-windows-ubuntu-bash

travis/setup_conda_windows.sh
```diff
@@ -6,12 +6,14 @@ if [[ $DEBUG == True ]]; then
...
 
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=4.5.4
+    MINICONDA_VERSION=4.6.14
 fi
 
 echo "installing miniconda3"
 choco install miniconda3 --params="'/AddToPath:1'" --version="$MINICONDA_VERSION";
-export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/:$PATH";
+export PATH="/c/tools/miniconda3/bin:$PATH"
```
djhoese: DEBUG run shows this as:

```
PATH='/c/tools/miniconda3/bin:/bin:/usr/bin:/c/tools/ruby25/bin:/c/Windows/system32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/c/Windows/System32/OpenSSH:/c/ProgramData/GooGet:/c/Program Files/Google/Compute Engine/metadata_scripts:/c/Program Files (x86)/Google/Cloud SDK/google-cloud-sdk/bin:/c/Program Files/Google/Compute Engine/sysprep:/c/Program Files/Docker:/c/ProgramData/chocolatey/bin:/c/Program Files/CMake/bin:/c/Program Files/Git/cmd:/c/Program Files/LLVM/bin:/c/Program 
```

But it still fails to find `conda`.

mwcraig:
```suggestion
export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/bin:$PATH"
```

djhoese: I don't think `bin` exists at all:

```++++ ls -1 /c/tools/miniconda3/bin
ls: cannot access '/c/tools/miniconda3/bin': No such file or directory```

Comment (mwcraig) [reply travis/setup_conda_windows.sh]: Ah, can you figure out where it is installing?

Comment (djhoese) [reply travis/setup_conda_windows.sh]: `conda.exe` seems to be in `scripts` as you have it. Conda does not create a `bin` directory on Windows so you can remove the second path.

I did just scripts and it found conda but it still didn't like using `conda activate base` without doing anything extra. I'll try source the bash profile and see what happens.

```

<details>

```

++++ ls -1 /c/tools/miniconda3/scripts
2to3-script.py
2to3.exe
activate
activate.bat
chardetect-script.py
chardetect.exe
conda-env-script.py
conda-env.exe
conda-script.py
conda.exe
deactivate
easy_install-script.py
easy_install.exe
idle-script.py
idle.exe
menuinst-script.py
menuinst.exe
pip-script.py
pip.exe
pydoc-script.py
pydoc.exe
pywin32_postinstall.py
pywin32_testall.py
wheel-script.py
wheel.exe
++++ conda init bash
no change     C:\tools\miniconda3\Scripts\conda.exe
no change     C:\tools\miniconda3\Scripts\conda-env.exe
no change     C:\tools\miniconda3\Scripts\conda-script.py
no change     C:\tools\miniconda3\Scripts\conda-env-script.py
no change     C:\tools\miniconda3\condabin\conda.bat
no change     C:\tools\miniconda3\Library\bin\conda.bat
no change     C:\tools\miniconda3\condabin\_conda_activate.bat
no change     C:\tools\miniconda3\condabin\rename_tmp.bat
no change     C:\tools\miniconda3\condabin\conda_auto_activate.bat
no change     C:\tools\miniconda3\condabin\conda_hook.bat
no change     C:\tools\miniconda3\Scripts\activate.bat
no change     C:\tools\miniconda3\condabin\activate.bat
no change     C:\tools\miniconda3\condabin\deactivate.bat
modified      C:\tools\miniconda3\Scripts\activate
modified      C:\tools\miniconda3\Scripts\deactivate
modified      C:\tools\miniconda3\etc\profile.d\conda.sh
modified      C:\tools\miniconda3\etc\fish\conf.d\conda.fish
no change     C:\tools\miniconda3\shell\condabin\Conda.psm1
modified      C:\tools\miniconda3\shell\condabin\conda-hook.ps1
modified      C:\tools\miniconda3\Lib\site-packages\xonsh\conda.xsh
modified      C:\tools\miniconda3\etc\profile.d\conda.csh
modified      C:\Users\travis\.bash_profile
==> For changes to take effect, close and re-open your current shell. <==
setting up common dependencies
++++ conda activate base
CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
If using 'conda activate' from a batch script, change your
invocation to 'CALL conda.bat activate'.
    $ conda init <SHELL_NAME>
Currently supported shells are:
  - bash
  - cmd.exe
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell
See 'conda init --help' for more information and options.
IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```

</details>

travis/setup_conda_windows.sh
```diff
@@ -6,12 +6,14 @@ if [[ $DEBUG == True ]]; then
...
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=4.5.4
+    MINICONDA_VERSION=4.6.14
 fi
 
 echo "installing miniconda3"
 choco install miniconda3 --params="'/AddToPath:1'" --version="$MINICONDA_VERSION";
-export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/:$PATH";
+export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/bin:$PATH"
+conda init bash
```
djhoese: Adding the following and removing `/c/tools/miniconda3/bin` from the above PATH got me a lot further.

```suggestion
conda init bash
source /c/Users/travis/.bash_profile
```

djhoese: After the above I then get:

```
++++++ export PATH
++++++ /c/tools/miniconda3/Scripts/conda.exe install conda
Collecting package metadata: ...working... done
Solving environment: ...working... failed
SpecsConfigurationConflictError: Requested specs conflict with configured specs.
  requested specs: 
  pinned specs: 
    - conda==4.7
Use '\''conda config --show-sources'\'' to look for '\''pinned_specs'\'' and '\''track_features'\''
configuration parameters.  Pinned specs may also be defined in the file
C:\tools\miniconda3\conda-meta\pinned.
```

djhoese: Found some old code I had in my project's travis that were manually removing the `defaults` channel. This was before strict channel priority was a thing. I'm removing it to see if that changes anything.

djhoese: I'm still getting the above error. I started talking to some folks on the conda-forge gitter for help. They are wondering if this is a bug. @mwcraig see https://gitter.im/conda-forge/conda-forge.github.io

djhoese:
Shouldn't the `source activate test` in `setup_dependencies_common.sh` be `conda activate test`?

travis/setup_conda_windows.sh
bsipocz: hmm, ideally this should not be miniconda, but that we download (see my failed attempt here, still need to sort out some path issue: https://github.com/astropy/ci-helpers/compare/master...bsipocz:travis_windows_latest_miniconda)

mwcraig:
Just a heads up: I’m away from keyboard for another ~2 hours

djhoese:
Newest update: I've updated all the environment scripts on my fork based on this branch to re-source bash profile instead of setting PATH to point to `bin`. This means that we can use `conda activate` properly:

```
bash miniconda.sh -b -p $HOME/miniconda
conda init bash
source ~/.bash_profile
```

I then ran in to a possible bug in conda that I'd like @mwcraig and @bsipocz to give me some feedback on. I mentioned it in the source code comments, but basically I get this:

```
++++++ export PATH
++++++ /c/tools/miniconda3/Scripts/conda.exe install conda
Collecting package metadata: ...working... done
Solving environment: ...working... failed
SpecsConfigurationConflictError: Requested specs conflict with configured specs.
  requested specs: 
  pinned specs: 
    - conda==4.7
Use '\''conda config --show-sources'\'' to look for '\''pinned_specs'\'' and '\''track_features'\''
configuration parameters.  Pinned specs may also be defined in the file
C:\tools\miniconda3\conda-meta\pinned.
```

To get around that in my own project's environment I removed the part of the code that writes the conda version of a pinning file. Now I'm getting a hanging terminal on some travis Windows because `_output=$($@ 2>&1)` in `retry_on_known_error` when the command fails. Still playing with stuff.

mwcraig:
@djhoese -- looks like you have made some good progress! I'll update this to source the bash_profile so that we can activate properly. If we don't then the first response we'll get to question will (understandably) be "please update to the currently correct thing".

Are you hitting the `SpecsConfigurationConflictError` only on Windows or on all platforms?

Let me know if I can help with the hanging terminal...

travis/setup_conda_windows.sh
```diff
@@ -6,12 +6,14 @@ if [[ $DEBUG == True ]]; then
...
 
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=4.5.4
+    MINICONDA_VERSION=4.6.14
 fi
 
 echo "installing miniconda3"
 choco install miniconda3 --params="'/AddToPath:1'" --version="$MINICONDA_VERSION";
-export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/:$PATH";
+export PATH="/c/tools/miniconda3/scripts:/c/tools/miniconda3/bin:$PATH"
```
mwcraig:
```suggestion
export PATH="/c/tools/miniconda3/scripts:$PATH"
```

djhoese:
I *think* the specs thing was only on Windows, but I honestly don't remember and wasn't keeping notes.

I'm still getting the hanging terminal. My branch is [here](https://github.com/djhoese/ci-helpers/tree/upgrade-conda-use-channel-strict) if you want to look. Last commit was to add a `set +e/-e` around the retry function. Have to go offline for a couple hours now.

mwcraig:
> I think the specs thing was only on Windows, but I honestly don't remember and wasn't keeping notes.

I haven't seen it on other platforms...

mwcraig:
Detailed notes below, but I'd recommend this: 

+ we drop some of the pretty old test cases: 
    - `Python 3.4`...released over 5 years ago. That fixes `1258.7`.
    - `pytest=2.7.3`...released over 4 years ago. Fixes `1258.11`.
    - `sphinx=1.4.1`...released over 3 years ago. Fixes `1258.12`.
+ Alternatively, just use pip to handle the really old cases.
+ Change `CHANNEL_PRIORITY=strict` to `flexible` unless there is some reason we really need `strict`. I think the old `true` is closer to meaning `flexible` than `strict` anyway. Fixes 1258.22, I think.

Going to sleep on fixes/investigation into 1258.4, read on for more detailed notes.

Notes on remaining failures:

+ [1258.4](https://travis-ci.org/astropy/ci-helpers/jobs/571544546) fails because the assertion at the end about Sphinx version fails. Not sure why, because  the requirement is `>1.6` and `2.1.2` is installed.
+ [ 1258.7](https://travis-ci.org/astropy/ci-helpers/jobs/571544549) fails because ~~no conda package exists for  astropy 2.0.14 on Python 3.4~~. Nope, it's complicated:
    - First, maybe drop PYTHON=3.4 cases because (from each time pip installs something in this run):  `DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).`?
    - The place where things go wrong is essentially [here](https://travis-ci.org/astropy/ci-helpers/jobs/571544549#L6906), where the wrong version of scikit-learn is installed. The hacky code I wrote to check for this output could work here too, but should really become a function at that point instead of copy/pasting it.
+ [1258.11](https://travis-ci.org/astropy/ci-helpers/jobs/571544553) fails because conda insists on installing a newer version of pytest than the pinned version (2.7.3). conda is failing because 2.7.3 was never packaged by conda on any standard channel and it predates conda-forge. Could fix this by immediately reverting to a pip install. Or, we could delete the case. The correct error is being raised by `setup_dependencies_common`.
+ [1258.12](https://travis-ci.org/astropy/ci-helpers/jobs/571544554) fails because of the conda-overrides-pinning bug [here](https://travis-ci.org/astropy/ci-helpers/jobs/571544554#L2665) when it installs sphinx. It may well fail because of other pinnings, didn't check.
+ [1258.22](https://travis-ci.org/astropy/ci-helpers/jobs/571544564) fails for a more interesting reason. It requires a pretty old `jinja2=2.8`, which is present on the anaconda `free` channel, so you might think that doing `restore_free_channel true` would work. When `free` is added it is a lower priority than the other default channels, **and `CHANNEL_PRIORITY=strict`** on this case. That priority means conda stops looking further when it gets to the first channel that has `pytest` in it, so adding the `free` channel won't help, which is my added code raises an error.

bsipocz:
> we drop some of the pretty old test cases:

    Python 3.4...released over 5 years ago. That fixes 1258.7.
    pytest=2.7.3...released over 4 years ago. Fixes 1258.11.
    sphinx=1.4.1...released over 3 years ago. Fixes 1258.12.


I'm in favour of these, but the tests need to be updated accordingly, they are rarely test these versions but instead some functionality or workaround.

bsipocz:
> Alternatively, just use pip to handle the really old cases

perfecly fine approach, too. The only exception is to get an old python conda installed to begin with, but I assume that still works?

travis/setup_dependencies_common.sh
```diff
@@ -140,7 +140,7 @@ fi
 # We pin the version for conda as it's not the most stable package from
 # release to release. Add note here if version is pinned due to a bug upstream.
 if [[ -z $CONDA_VERSION ]]; then
-    CONDA_VERSION=4.5.12
+    CONDA_VERSION=4.7
```
djhoese: When using in a pinned file I think this needs to be `4.7.*` otherwise the solver tries to find something with the exact version "4.7" which it will never find.

travis/setup_conda_osx.sh
```diff
@@ -28,6 +28,8 @@ fi
 wget https://repo.continuum.io/miniconda/Miniconda3-${MINICONDA_VERSION}-MacOSX-x86_64.sh -O miniconda.sh
 bash miniconda.sh -b -p $HOME/miniconda
 export PATH="$HOME/miniconda/bin:$PATH"
+conda init bash
+conda activate base
```
djhoese: I've noticed some strange behavior or warnings and have now changed the osx/linux scripts to do:

```
# Create .conda directory before install to workaround conda bug
# See https://github.com/ContinuumIO/anaconda-issues/issues/11148
mkdir $HOME/.conda
bash miniconda.sh -b -p $HOME/miniconda
$HOME/miniconda/bin/conda init bash
source ~/.bash_profile
conda activate base
```

This changes the first call to conda so it uses an exact path. This stops ci-helpers from having to modify the PATH and then immediately re-source the bash profile which will do it anyway.

djhoese:
Another update: I got OSX and Linux working for my project and got Windows to stop hanging (with a hack) but it then failed because of a buggy conda-forge package (see https://github.com/conda-forge/proj.4-feedstock/issues/56).

The main issue `ci-helpers` has (in addition to the old versions mentioned above) is that it sometimes calls `conda install` from a subshell. With the way conda now uses activate/deactivate scripts this won't work because we need the environment variable changes in the subshell to effect the current environment. `conda install` will automatically reactivate an environment when it's done so any new activation scripts are executed. For some reason this is working well enough on Linux/OSX but Windows just hangs.

Anyone know how to collect the output of a command but run it in the current shell environment?

djhoese:
Ok, last update until changes are made by @mwcraig in this PR. My most recent changes (summarized [here](https://github.com/djhoese/ci-helpers/compare/bebe7cc..b9b90c8)) get me a working environment for my pyresample project *except* for the buggy conda-forge PROJ4 Windows package. Still waiting on a response for that.

The major change had two parts:

1. Add `set +e/-e` around the `retry_on_known_error` functions execution of the command to be run. I think this stops the hanging in Windows and some other platforms where any failure on conda's scripts (like, I think, when it fails to solve in the shortcut fashion and has to unfreeze the environment and try again).
2. Write output to a temporary file instead of `$(...)` subshell. This makes it so the conda script runs *inside* the current shell and can effect our current PATH and other environment variables. Since we want to collect the output of the command we need to write it somewhere without using a subshell, that's why there is a temporary file. This may not be needed but depends on  what happens after the conda-forge PROJ.4 library is fixed. Right now my failures are because things aren't being activated because the activation scripts are missing for proj4.

travis/setup_dependencies_common.sh
mwcraig: Makes sense; at some point soon we need to change all of the pinnings to ranges, e.g. `>=4.7,<4.8` because I think the `.*` style is going away at some point (There have been deprecation warnings in the logs about it).

djhoese:
The conda-forge proj package was fixed, but I have another idea why `conda install $CONDA_DEPENDENCIES` keeps failing or rather hanging. The scripts capture all output and since the default behavior for ci-helpers is to force a version of numpy in addition to packages already installed (instead of installing everything in one go which would be faster), travis doesn't see any activity for 10+ minutes and times out.

mwcraig:
Thanks for all of the suggestions, @djhoese! I think I made your changes propperly in https://github.com/astropy/ci-helpers/pull/405/commits/3cb4ac1a6d647fd225851580e55c75f6a71cdb20

@bsipocz -- I tried to update the travis matrix in https://github.com/astropy/ci-helpers/pull/405/commits/6c8ee509e515f4ba12f17e360ce7c2f3bcbe1296; please have a look when you get a chance

.travis.yml
```diff
@@ -74,136 +90,185 @@ matrix:
...
+        # JOB .26
         - os: linux
           env: SETUP_CMD='test --numprocesses'
                TEST_CMD='py.test --numprocesses 4 test_env.py'
 
+        # JOB .27
         - os: linux
           env: SETUP_CMD='test --open-files'
 
+        # JOB .28
```
bsipocz: this job seems to be unnecessary, and can be removed altogether

.travis.yml
```diff
@@ -74,136 +90,185 @@ matrix:
...
         # -> Have package dependency that matches partially another package
         #    (pytest vs pytest-cov) when falling back to pip install
         # -> testing whether version numbers are respected by pip install fallback
-        - os: linux
-          env: SETUP_CMD='test --coverage' PYTHON_VERSION=3.4
+        # The following versions were updated in 2019 because the old versions
+        # were leading to issues because anaconda deprecated the old free channel:
+        #
+        #    PYTHON_VERSION 3.4 → 3.6
+        #    NUMPY_VERSION 1.10 → 1.16
```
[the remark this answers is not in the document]
mwcraig: So this one is intended to check whether version numbers are respected by pip when the pip fallback is used? Or is the idea that we want the conda solve to fail for sure by using the combination of old numpy and newer scikit-learn that conda can't solve?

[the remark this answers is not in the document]
bsipocz: yes, early on there was a bug that the pip install fallback didn't take into account the version pinning, e.g. `SCIKIT_LEARN_VERSION`. So this job needs a combination that has a package with a version that is not conda installable, but pip needs to respect the version being set in the env variable.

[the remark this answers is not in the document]
mwcraig: ok; somewhat unrelated question: is it ok to write a pytthon script that is used by `setup_dependencies_common.sh`? I need a way to hack around a conda bug and it would be much easier to do the processing in python than in bash...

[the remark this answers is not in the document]
bsipocz: yes, I suppose that's OK. We started this as a bash script as it was a simple ~50 line thing, then it started to grow organically to a monster. (Also I kind of like the bash hacking, but now parts of this are so much out of control that bugs remain undetected for quite a while).

travis/setup_dependencies_common.sh
```diff
@@ -37,14 +37,51 @@ function retry_on_known_error() {
...
-        _output=$($@ 2>&1)
-        _exitval="$?"
+        # Execute the wrapped command and get its unified output.
+        # This command needs to run in the current shell/environment in case
+        # it sets environment variables (like 'conda install' does)
+        #
+        # Wow, bash is fussy about order. Redirection to a file needs to
+        # happen *before* the redirection of stderr to stdout
+        $@ 2>&1 | tee $_tmp_output_file
+        _exitval="${PIPESTATUS[0]}"
```
mwcraig: @djhoese -- you should also change `$?` to this, otherwise `tee` is the last process and always succeeds (or so I'm told by stack overflow 🤣 )

bsipocz:
@mwcraig - we should not make confession in file names 😆

travis/setup_dependencies_common.sh
djhoese: Ah neat. I will probably rebase any of my changes on your branch to double check that my project can pass with what's been done here so far.

travis/setup_dependencies_common.sh
```diff
@@ -525,8 +661,14 @@ if [[ $SETUP_CMD == *build_sphinx* ]] || [[ $SETUP_CMD == *build_docs* ]]; then
...
             PIP_PACKAGE_VERSION=$(awk '{print $2}' $PIN_FILE)
+            # Debugging....
+            echo "WHAT IS GOING ON HERE (TAKE 2)"
+            conda info -a
+            conda config --show
+            conda list
+            cat $PIN_FILE
             if [[ $(echo $PIP_PACKAGE_VERSION | cut -c 1) =~ $is_number ]]; then
-                PIP_PACKAGE_VERSION='=='${PIP_${package}_VERSION}
+                PIP_PACKAGE_VERSION='=='${PIP_PACKAGE_VERSION}
```
mwcraig: @bsipocz -- this was a bug, right? If I try that substitution locally I get `-bash: ${PIP_${package}_VERSION}: bad substitution` (and was seeing same error on travis)

mwcraig:
Progress, we are down to two failures:

+ Job .7 has a green light but [actually fails](https://travis-ci.org/astropy/ci-helpers/jobs/573198575#L6668). Not sure what is going on here, so I'll add one more debugging line for this case before trying to simplify the requirements in the job.
+ Job .11 fails because  the [version of maplotlib installed, `1.5.1`](https://travis-ci.org/astropy/ci-helpers/jobs/573198579#L4987) is not what I expected from the spec I gave it, `matplotlib<=2.2.0`, though the installed version is *consistent* with the spec it was given. WIll try making the spec `>2.0,<=2.2.0` to see what happens.

My thinking is to get these two working, then perhaps merge this, then start pruning down the build matrix based on @bsipocz's comments above.

mwcraig:
🎉 🎈 🎂 🙌

mwcraig:
Now that the celebration is done, note that job **.7** still fails: https://travis-ci.org/astropy/ci-helpers/jobs/573474730#L6859

djhoese:
@mwcraig I've discovered another issue with the way ci-helpers is doing things. It can in general be resolved by deactivating/reactivating the environment, but I'm wondering if we should work harder to make that unnecessary. The main issue is that running a bash function in a pipe (not sure on chained conditionals) will cause it to run in a subshell. This means none of the conda activation scripts will be run in the current sourced environment. We could try using `{ }` instead of `( )` to group commands which should make them run in the current shell. Not sure the best way to solve this overall though.

djhoese:
See https://github.com/djhoese/ci-helpers/commit/a004105091a7cd81eb502e251131c59d893ee8ed

I can't test it currently because gdal is broken on conda-forge again (that package will be the death of me).

mwcraig:
> The main issue is that running a bash function in a pipe (not sure on chained conditionals) will cause it to run in a subshell.

Interesting...so do pipes *always* run in a subshell? That takes me back to "learning more about bash than I ever wanted to know." 😿 

Looks like yes according to google: https://stackoverflow.com/questions/5760640/left-side-of-pipe-is-the-subshell.

So what if we dump the conda init stuff into `.bashrc`? By default `conda init` modifies `.bash_profile`, which is only run for login shells. On my local machine it seems like that comes  closer to activating the base environment when I start a subshell.

djhoese:
So the above commit fixes some things and changing the retry method to use this:

```
$@ 2>&1 > >(tee $_tmp_output_file)
_exitval="$?"
```

I *think* this keeps the running command in the current shell. I was going to test it...and then conda-forge's gdal package broke again. Waiting 50 minutes to try again (for anaconda.org index to refresh).

djhoese:
Oh wait, the retry method uses a while loop which I think is itself a subshell...ugh.

Anyway, I'm not sure I understand your bashrc suggestion. The issue isn't that conda isn't init'd before we use it, it's that `conda install` wants to modify environment variables that need to exist when `python` is called later (during the tests). We could just reactivate the environment when all is said and done.

djhoese:
It worked!

https://travis-ci.org/pytroll/pyresample/jobs/573960541

mwcraig:
> Anyway, I'm not sure I understand your bashrc suggestion. The issue isn't that conda isn't init'd before we use it, it's that conda install wants to modify environment variables that need to exist when python is called later (during the tests). We could just reactivate the environment when all is said and done.

Ah...so the issue is not what conda environment is running in the subshell (we could make sure it is test), it is what/the value of  environment variables that are set in the bash shell in which the python process is running?

mwcraig:
> It worked!

Can you point me to the commit(s) I should look at for the changes you made (if there are any beyond the one you linked to above, https://github.com/djhoese/ci-helpers/commit/a004105091a7cd81eb502e251131c59d893ee8ed and the change to retry above?

Not 100% sure I will have a chance to update the PR today, but will try...

djhoese:
Yes, if I get your question. Example:

```
conda update conda
conda create -n test python=3.7 ...
conda activate test  # activation scripts executed which set/manipulate environment variables
conda install pyproj  # whole environment is re-activated including PROJ's which sets PROJ_LIB
python my_pyproj_based_script.py
```

If the `conda install pyproj` command happens in a subshell then the reactivation doesn't happen in the current environment which means `PROJ_LIB` is never set (in addition to whatever else changed from the installed packages).

djhoese:
There are two commits after the one I referenced. One was unsuccessful and the other is https://github.com/djhoese/ci-helpers/commit/031de5d47424bb90b9f72467f19e16465bddc602 but is based on the unsuccessful one. Here's the full list: https://github.com/djhoese/ci-helpers/commits/upgrade-conda-use-channel-strict

I'll try making a new branch rebased on this PR's branch and see how it goes. I'd also like to take a stab at appveyor and see what type of pain I can bring on myself.

mwcraig:
I made some progress on appveyor here: https://github.com/astropy/ci-helpers/pull/403

I think the power she’ll retry command is broken and have no clue how to fix it. It might be better to simply scrap the retry on appveyor...an HTTP error followed by a manual reset of a job isn’t that hard anyway.

djhoese:
Ok so the plan is then to make a few last changes to this PR. Merge it. Then merge master in to #403, then see what else is broken? I see there are still some errors there. I wonder if any of the lessons learned here can be applied there.

mwcraig:
I think the plan would be to merge this, then rebase #403 (or maybe start fresh, honestly). Once we are happy with how this one is working I'll squash to make it easier to review (I don't have merge rights in this repo, so we'll need to get a merge form the folks who do).

bsipocz:
Thank you both for working on this. I'll try to review this asap, but can't promise more that it'll be on this week (as I'm travelling and there a couple of other, urgent, fires I need to attend first).

mwcraig:
@djhoese -- I've incorporated your changes in #414 here, and rebased/squashed to simplify the history (which was getting a little convoluted between the two branches).

We'll see what happens 🤞

djhoese:
@mwcraig Another small fix to your hack_version_number.py script: https://github.com/djhoese/ci-helpers/commit/64b1e4de8f9c287fde6a4ec9caef214de3123e06

Without this it was overwriting `conda install conda ...` with `conda=4.7.11 install conda=4.7.11 ...`. Testing it with my own project now.

README.md
```diff
@@ -143,10 +143,10 @@ environment variables
   be ``SCIKIT_IMAGE_VERSION``). If specified it will override any version
   number limitations listed in ``CONDA_DEPENDENCIES``.
 
-* ``CONDA_CHANNEL_PRIORITY``: can be set to ``True`` or ``False``, and
+* ``CONDA_CHANNEL_PRIORITY``: can be set to ``strict``, ``flexible`` or ``disabled``, and
```
bsipocz: what will happen to repos that used `True` and `False` before?

README.md
```diff
@@ -143,10 +143,10 @@ environment variables
...
   affects the ``channel_priority`` conda setting (as discussed
-  [here](http://conda.pydata.org/docs/channels.html). The default is
-  ``False``.
+  [here](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html). The default is
+  ``false``.
```
bsipocz: bool options are not listed above, should this be `'disabled'`?

travis/setup_conda.sh
```diff
@@ -7,6 +7,7 @@
 #
 # The present script was added later.
 
+export DEBUG="True"
```
bsipocz: I suppose this was a debug change that needs to be removed

travis/setup_conda_windows.sh
```diff
@@ -6,12 +6,14 @@ if [[ $DEBUG == True ]]; then
 fi
 
 if [[ -z "${MINICONDA_VERSION}" ]]; then
-    MINICONDA_VERSION=4.5.4
+    MINICONDA_VERSION=4.6.14
```
bsipocz: is 4.7 not working with windows?

.travis.yml
```diff
@@ -13,26 +13,32 @@ env:
         - SETUP_CMD='test'
         - TEST_CMD='py.test test_env.py'
         - PYTHON_VERSION=3.6
+        - DEBUG=True
```
bsipocz: If this is left as global here, we better add a few `DEBUG=False` ones to the various jobs below

.travis.yml
```diff
@@ -46,14 +52,23 @@ matrix:
...
         # -> Testing whether pip install fallback is working for sphinx/matplotlib in the docs builds
         # -> Test we have a recent pip version, the newest on the defaults channel
         #    is 10.0.1 while the latest is 18.0 at the time of writing this test.
+        # -> The latest version of pip as of 2019-08-15 on default channels is 19.2.2
+        #    for Python >=3.6, so bump the pip version check to minimum of 19.
+        # Updated these versions because of availability of packages on the default
+        # channels:
+        #
+        #    PYTHON_VERSION 3.5 → 3.6
+        #    pip minimum version >17 → >18
```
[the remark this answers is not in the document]
bsipocz: these comments can be removed

.travis.yml
```diff
@@ -74,136 +90,186 @@ matrix:
...
-        - os: linux
-          env: SETUP_CMD='build_sphinx' SPHINX_VERSION=1.4.1 DEBUG=True
-               MATPLOTLIB_VERSION=1.4 DEBUG=True PYTHON_VERSION=3.5 PYTEST_VERSION=3.0
-               TEST_CMD='python -c "import sphinx; assert sphinx.__version__==\"1.4.1\"; import pytest; pytest.__version__.startswith(\"3.0\"); import matplotlib; assert matplotlib.__version__=="\""1.4.3"\"'
-
+        #
+        # Changed version numbers in 2019 because older versions are no longer
+        # available in default conda channels:
+        #
+        #    matplotlib 1.5.0 → 2.2.0
```
bsipocz: no need for comments in the file, having the updates and these info in the commit message is enough

travis/setup_dependencies_common.sh
```diff
@@ -37,25 +37,69 @@ function retry_on_known_error() {
...
-        _output=$($@ 2>&1)
+        # Execute the wrapped command and get its unified output.
+        # This command needs to run in the current shell/environment in case
+        # it sets environment variables (like 'conda install' does)
+        #
+        # tee will both echo output to stdout and save it in a file. The file
+        # is needed for some error checks later. Output to stdout is needed in
+        # the event a conda solve takes a really long time (>10 min). If
+        # there is no output on travis for that long, the job is cancelled.
+        set +e
```
bsipocz: wouldn't this be littering the logs too much?

travis/setup_dependencies_common.sh
```diff
@@ -37,25 +37,69 @@ function retry_on_known_error() {
...
         fi
+
         # The command errored, so let's check its output for the specified error
         # strings:
         if [[ $_n_retries -lt $RETRY_MAX ]]; then
             # If a known error string was found, throw a warning and wait a
             # certain number of seconds before invoking the command again:
             for _error in $RETRY_ERRORS; do
-                if [ -n "$(echo "$_output" | grep "$_error")" ]; then
+                if [ -n "$(grep "$_error" "$_tmp_output_file")" ]; then
```
bsipocz: weird that the `""` are needed around the file name, but if it works 🤷‍♀

travis/setup_dependencies_common.sh
```diff
@@ -313,7 +374,85 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
...
+    #
+    #     The shell does not exit if the command that fails is ...part of the
+    #     test in an if statement...
+    #
+    # Use tee to print output to console and to file to avoid travis timing out
+    _tmp_output_file="tmp.txt"
+    # do not exit on failure of the dry run because pip fallback may succeed
+    set +e
+
+    conda install --dry-run $CONDA_DEPENDENCIES > >(tee $_tmp_output_file) 2>&1
```
bsipocz: Out of curiosity what does `> >` mean?

travis/setup_dependencies_common.sh
```diff
@@ -313,7 +374,85 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
...
+    # 'grep' returns non-zero exit status if no lines match.
+    if [[ ! -z $(grep "conflicts with explicit specs" $_tmp_output_file) ]]; then
+        echo "restoring free channel"
+        # Restoring the free channel only helps if the channel priority
+        # is not strict, so check that first. If it is strict, fail instead
+        # of changing the solve logic.
+
+        # ...but only if the free channel is not ruled out by strict
+        # channel priority
+        if [[ ! -z $CONDA_CHANNEL_PRIORITY && $CONDA_CHANNEL_PRIORITY == strict ]]; then
```
bsipocz:
```suggestion
        if [[ $CONDA_CHANNEL_PRIORITY == strict ]]; then
```

bsipocz:
I didn't find any issues, so will merge once travis is passing again (haven't looked into it why it's failing, feel free to remove jobs that became hopeless over the time).

README.md
mwcraig: `True` → `flexible`
`False` → `disabled`

travis/setup_conda_windows.sh
mwcraig: It wasn't available on chocolatey a 1.5 months ago, but does appear to be now: https://chocolatey.org/packages/miniconda3

.travis.yml
mwcraig: We can remove it once this is ready to merge

travis/setup_dependencies_common.sh
mwcraig: I think `set +e` just tells the script not to exit if it hits an error without affecting the verbosity of the output.

The `cat` below this could be  wrapped in an "if DEBUG" but I think it is better if the output of the conda command shows up in the log.

travis/setup_dependencies_common.sh
mwcraig: Probably not *needed*, but they do no harm and allow for the case that someone wants to use spaces in the temporary output file name.

travis/setup_dependencies_common.sh
mwcraig: I know so much more arcane bash now than I did before; this bit of it is thanks to @djhoese. It turns out that `>(command)` pipes stdout into `command`; its partner `<(command)` connects the output of command to stdin. 

At least I think I have that right....

mwcraig:
@bsipocz -- please don't merge quite yet; there is still a major bug I found when I looked the "allowed failure" case. Right now `retry_on_known_failure` tries a `conda install`, checks for a pinning failure, then re-runs `conda install` with a command line pin.

The critical error here is that the `conda install` may well have installed the *wrong* version of the package, so its actions need to be rolled back. Fortunately, a little reading last night turned up `conda list --revision` and `conda install --revision`, so I think a rollback will be relatively straightforward.

travis/setup_dependencies_common.sh
djhoese: So this is known as "process substitution" iirc and I only found it because of trying to work on this issue. The idea is that most flavors of bash/sh will put any piped commands (ex. `one_cmd | two_cmd | three_cmd | tee tmp_file`) in a subshell. Normally, for most bash scripts, this isn't a huge deal, but for conda it is because the `conda` shell command is modifying `PATH` and running other scripts so that they effect our current environment. If they were run in a subshell then none of those changes take effect.

So what is happening is the `conda install` command is running in the current shell and then piping its output to a new process (which is running `tee`).

The other option would be to deactivate and reactivate conda every time we run a conda command in these setup scripts.

travis/setup_dependencies_common.sh
bsipocz: oh, I'm old school, spaces in file names? 😱

bsipocz:
@mwcraig - for sure, travis needs to be green before merging, and indeed that allowed to fail should actually fail (as still there is no way to set a job as an expected failure).

mwcraig:
Based on https://github.com/conda/conda/issues/9055#issuecomment-518756256 I think we should stick with 4.6 on travis/windows for now....

mwcraig:
@bsipocz -- assuming this passes, I think it is ready to go

mwcraig:
The job that is supposed to fail is failing in the way it should

travis/hack_version_numbers.py
```diff
@@ -32,9 +32,10 @@
         package = match.group(1)
         versions[package] = "=" + match.group(2)
 
-new_args = []
+# We never want to add a spec to the "conda install" part of the command
```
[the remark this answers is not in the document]
mwcraig: There might be additional flags, but that is ok, for the reason you point out. The problem had been that if conda had a spec, e.g. `4.7.11`, then  `conda install conda` got changed to `conda=4.7.11 install conda=4.7.11`.

bsipocz:
Thank you @mwcraig and @djhoese! And I'm sorry it took so long to get this merged in.

bsipocz:
Any README updates are welcome. Size wise it's getting out of hand, so it may very well be a time to restructure.

A real solution to the `source activate`/`conda activate` issue could be to open batch PRs for repos that use ci-helpers. We have some machinery for automated PRs in astropy, but I don't have any time before the weekend to get the bits and pieces together.

PR #407, opened 2019-08-05
Ensure PEP 8 runs for [docs only] directive

### LINKED ISSUE #404: Should [docs only] skip PEP 8 check too?
Currently, when `[docs only]` directive is given, everything but the doc build is skipped. However, why not check for PEP 8 anyway?

pllim:
Fix #404

### DIFF
--- a/travis/setup_conda.sh
+++ b/travis/setup_conda.sh
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ] && [MAIN_CMD =~ pycodestyle|flake8|pep8]]; then
+    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
         # we also allow the style checkers to run here
         echo "Only docs build was requested by the commit message, exiting."
         travis_terminate 0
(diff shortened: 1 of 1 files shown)


travis/setup_conda.sh
```diff
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake|pep8 ]]; then
+    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ] && [MAIN_CMD =~ pycodestyle|flake8|pep8]]; then
```
bsipocz:
```suggestion
    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
```

bash is sensitive to white spaces around brakets

bsipocz:
@pllim - I've cancelled CI as there was a syntax error, otherwise it looks good to me.

bsipocz:
(I don't understand why it keeps saying that I don't have permissions to push to this repo, it seems that you need to accept the suggested changes yourself. )

pllim:
Ops, thanks!

pllim:
Three jobs failed, all seem related to `nomkl`. Not sure how to fix them. Please advise. Thanks!

bsipocz:
Interesting, not sure what has changed with the versions so now the combination is not conda compatible any more. `nomkl` is not on pip, so the fallback naturally fails there. 
That is a bug, but the tests are there to test the working `nomkl` case on conda, so I suppose we need to pin some versions.

bsipocz:
I recall @astrofrog mentioned something recently about `mkl`/`nomkl`, maybe he has some hints for here, too.

bsipocz:
What the heck. I go ahead and merge this, it's better than it was before and no point in waiting with the merge until #408 is fixed (as that effect master already).

bsipocz:
Thanks @pllim

PR #409, opened 2019-08-07
Leave nomkl alone if numpy is not defined by NUMPY_VERSION

### LINKED ISSUE #408: Do not pip fall back for `nomkl`
`nomkl` is a conda workaround, we should not try to pip install it in case of version incompatibility

bsipocz:
To fix #408


bsipocz:
Nice greem travis once again 🎉

PR #414, opened 2019-08-20
Use conda activate in travis environments

djhoese:
This adds a few changes to #405. Mainly it uses `conda activate` to activate conda environments instead of `source activate` which is now deprecated. Care was taken to keep `conda install` commands running in the same shell environment as the main script so they could be sourced properly. See discussion in #405 for more info.


djhoese:
@mwcraig I caught some big issues with the way I was doing things in my last commit here. Mainly that the process substitution I was using wasn't properly giving stderr to `tee`. Commands need to look like:

```
conda install --dry-run $CONDA_DEPENDENCIES > >(tee $_tmp_output_file) 2>&1
```

Note that I also moved the `--dry-run` commands outside their pipelines so we could pass the output to tee and avoid travis timeouts. Lastly, I have one failing travis test left because of an incompatible dependency list: https://travis-ci.org/astropy/ci-helpers/jobs/574445926

I'm wondering if this is after a few updates from conda and whether your own PR would fail too for this environment?

bsipocz:
@djhoese @mwcraig - #405 and this have diverged. Please give me a sign which one and when to review/merge.

djhoese:
I'm 90% sure this PR's changes have been manually ported to #405 (I would test something here, I comment on 405, @mwcraig updates, I test again). Let's close this and make #405 the PR to be merged.

PR #415, opened 2019-08-20
Fix appveyor builds with new conda

djhoese:
Successor to #403

PR #416, opened 2019-08-21
Test more modern appveyor setup

djhoese:
Built on #415, this is my attempt at getting rid of all of the extra PATH and activation hackery that was needed. Probably won't work, but let's see what happens.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ environment:
         PYTEST_VERSION: "=3"
         TEST_CMD: "python -c \"import pytest; assert int(pytest.__version__[0])<4\""
 
-      - PYTHON_VERSION: "3.5"
+      - PYTHON_VERSION: "3.7"
         NUMPY_VERSION: "1.13"
         ASTROPY_VERSION: "lts"
 
@@ -48,7 +48,7 @@ environment:
         # on Numpy to make sure that Numpy dev is used (because conda will also
         # install Numpy stable)
 
-      - PYTHON_VERSION: "3.4"
+      - PYTHON_VERSION: "3.7"
         NUMPY_VERSION: "1.9"
 
         # We deliberately test with Numpy 1.13 to check that installing Astropy
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,7 @@ environment:
         CONDA_DEPENDENCIES: "scipy glymur"
         CONDA_CHANNELS: "conda-forge astrofrog"
         PIP_DEPENDENCIES: "requests pyparsing"
-        CONDA_CHANNEL_PRIORITY: "True"
+        CONDA_CHANNEL_PRIORITY: "flexible"
         DEBUG: "True"
 
       - PYTHON_VERSION: "3.6"
@@ -40,7 +40,7 @@ environment:
         ASTROPY_VERSION: "stable"
         CONDA_DEPENDENCIES: "matplotlib h5py"
         PIP_DEPENDENCIES: "astrodendro"
-        CONDA_CHANNEL_PRIORITY: "False"
+        CONDA_CHANNEL_PRIORITY: "disabled"
         SUNPY_VERSION: "dev"
         DEBUG: "True"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ environment:
         PYTEST_VERSION: "=3"
         TEST_CMD: "python -c \"import pytest; assert int(pytest.__version__[0])<4\""
 
-      - PYTHON_VERSION: "3.7"
+      - PYTHON_VERSION: "3.6"
         NUMPY_VERSION: "1.13"
         ASTROPY_VERSION: "lts"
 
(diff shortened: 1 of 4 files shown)


djhoese:
@mwcraig FYI I got the appveyor tests to work by removing all the hacks from `appveyor.yml` and adding `conda init cmd.exe` to the install script. My latest commit is checking if the test environment actually needs to be re-activated.

djhoese:
Last commit didn't work, the environment does need to be activated in cmd.exe to work.

djhoese:
@bsipocz I would say this is the best solution and replaces #415 and therefore #403 also.

djhoese:
Still need to update the README, but let me know what you think.

appveyor.yml
```diff
@@ -31,23 +32,23 @@ environment:
...
         SUNPY_VERSION: "dev"
         DEBUG: "True"
 
         # For the Numpy dev build, we also specify a conda package that depends
         # on Numpy to make sure that Numpy dev is used (because conda will also
         # install Numpy stable)
 
-      - PYTHON_VERSION: "3.4"
+      - PYTHON_VERSION: "3.7"
         NUMPY_VERSION: "1.9"
```
mwcraig: I'm surprised the combination Python 3.7/numpy 1.9 works!

mwcraig: Surprised, but numpy 1.9 is available on python 3.7

appveyor.yml
```diff
@@ -91,4 +89,6 @@ install:
 build: false
 
 test_script:
+  - "conda info -a"
+  - conda list
   - "%CMD_IN_ENV% %TEST_CMD%"
```
mwcraig: Let's go ahead and get rid of the `%CMD_IN_ENV%` thing as long as we are cleaning up. That magic isn't necessary for python 3.5 or later.

```suggestion
```

djhoese:
@mwcraig @bsipocz Let me know if any of the existing appveyor environments should be updated and if I should still implement the suggested change(s) above from @mwcraig.

appveyor.yml
bsipocz: I would keep as long as it works.

bsipocz:
@djhoese - Apparently I cannot accept the code change suggestion above. So if you would do that then I think this ready to go.

appveyor.yml
djhoese: This suggestion shows up as removing the whole line (right?), but it sounds like @mwcraig is saying only remove the `%CMD_IN_ENV%` part. How should I change this?

bsipocz: oh, you're right. That case you cannot accept the suggestion, but just  do the edit yourself.

bsipocz:
OK, rebase is done here, too. Merging away once all green!

test_env.py
```diff
@@ -242,7 +242,7 @@ def test_regression_mkl():
 
 def test_conda_channel_priority():
 
-    channel_priority = os.environ.get('CONDA_CHANNEL_PRIORITY', 'False')
+    channel_priority = os.environ.get('CONDA_CHANNEL_PRIORITY', 'disabled')
```
bsipocz: This pretty badly fails the travis tests so I'm adding a workaround for now that should be removed in #405

bsipocz:
I'm not sure what the issue is with the LTS tests.

bsipocz:
I've cancelled travis

djhoese:
The `dev` version of matplotlib (current master branch) switched to using PIL instead of libpng directly (from what I understand). The change was made very recently [here](https://github.com/matplotlib/matplotlib/commit/370e9a2d5d9e637abc90b3270d368642c69f66c6). This added `pillow` as a dependency so I'm guessing whatever stuff ci-helpers is doing to handle the dev version of matplotlib doesn't include installing the differing dependencies (pillow is not a dependency on the conda-forge package **yet**).

bsipocz:
@djhoese  - thanks for the heads up. I plan to switch off to use their nightly builds for the dev versions, that might resolve the dependency issues (though probably it won't, I don't know). Anyway, these are issues on master, too, so I'm happy to merge whenever you're happy with this (preferably the appveyor tests should pass, but they are so terribly broken atm on master that I'm not picky and would merge this as is if that's the preference).

djhoese:
Looks like the version of pytest on the first failing appveyor job is really old (3.3.2) and is using a new version of the `attrs` library where the `convert` keyword is now `converters` and is causing an error. I'm not sure if that is the older numpy version causing the old pytest to be installed.

bsipocz:
Yeah, and I'm messing around with the "stable" numpy version test. It's actually doing what it suppose to do, just the test is bad.

bsipocz:
Thanks, good investigation. Now it should really, really work.

bsipocz:
I give up, really. Have to make another PR with updating the astropy  versions, will try to hack around with the remaining one appveyor failure there. It's a test issue, so I'm very happy to be able to merge this now and declare appveyor fixed.

Thank you both a zillion times for the fix and your patience with my very slow turn-around time.

Issue #417, opened 2019-09-17
Revise  [docs only] for repos that use circleCI for their docs build


bsipocz:
Jobs shouldn't be run tests here:
https://travis-ci.org/astropy/astropy/builds/586132358?utm_source=github_status&utm_medium=notification

Also, now that many projects are moving their docs builds to circleCI from travis, `[docs only]` could be made a bit more clever, e.g. if there isn't even one docs job in the travis config the hole build should be cancelled without going into the individual jobs.

pllim:
Yeah, something like not right with this logic somehow but my bash-fu is not strong enough...

https://github.com/astropy/ci-helpers/blob/e1bddf10a07e88f571ea7e382051b5cab6fc4ee4/travis/setup_conda.sh#L49-L54

It looks as if `[docs only]` has *no effect* now; i.e., all the Travis jobs run regardless.

pllim:
Maybe I broke it in #407 . :sob:

pllim:
I think this is now fixed?

pllim:
Oh, wait... Haven't done "if there isn't even one docs job in the travis config the hole build should be cancelled without going into the individual jobs"... BUT do we really want that? I thought we still want PEP 8 job to run regardless.

bsipocz:
There is a pep8 test on circleCI now, too, so it's all good to cancel the whole of travis.

pllim:
But, only the core does that, not everyone else who uses `ci-helpers`?

bsipocz:
oh, yes, you're right regarding that.

pllim:
Maybe the title of the issue should at least be updated, as `[docs only]` is no longer broken.

PR #418, opened 2019-09-17
BUG: Fix broken [docs only] directive

pllim:
@jhunkeler pointed out my sins to me in #407 .

I'll find a way to test this before merge...

Edit: Fix the bug part of #417, keeping that issue open until the enhancement part is investigated.

### DIFF
--- a/travis/setup_conda.sh
+++ b/travis/setup_conda.sh
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
+    if [[ ! $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ ! $MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
         # we also allow the style checkers to run here
         echo "Only docs build was requested by the commit message, exiting."
         travis_terminate 0
--- a/travis/setup_conda.sh
+++ b/travis/setup_conda.sh
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if [[ ! $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ ! $MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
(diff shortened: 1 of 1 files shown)


bsipocz:
That was on my sloppy review, sorry.

travis/setup_conda.sh
```diff
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
+    if [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ $MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
```
bsipocz: why remove the `!`? also the second half of the expression should be `||` rather than `&&`.

pllim: Argh... My brain kept interpreting `=~` as `!=`. >.<

bsipocz:
As for tests, I'm not sure, add `[docs only]` to all the commit messages here, and we can double check whether the irrelevant ones are skipped or not.

travis/setup_conda.sh
```diff
@@ -47,7 +47,7 @@ if [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${TR_SKIP}") ]]; then
     echo "Travis was requested to be skipped by the commit message, exiting."
     travis_terminate 0
 elif [[ ! -z $(echo ${COMMIT_MESSAGE} | grep -E "${DOCS_ONLY}") ]]; then
-    if ! [[ $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
+    if [[ ! $SETUP_CMD =~ build_docs|build_sphinx|pycodestyle|flake8|pep8 ]] && [[ ! $MAIN_CMD =~ pycodestyle|flake8|pep8 ]]; then
```
pllim: Is it better like this?

bsipocz: I prefer the other way as it's clearer what to negate, but I'm ok with this way, too if it works. (e.g. I'm not sure about the precedent order for `!` vs `=~`).

pllim: I think the CI for this PR is skipping the right stuff so far? Though I can't tell whether the doc build log means it is skipped or just it found nothing to build. The message difference is very subtle to me.

bsipocz: I'll have a look. Basically for the skipped jobs there shouldn't be any dependencies installed, the skip should happen before the script goes into parsing and installing anything.

bsipocz: It indeed seems to work, but could you also add `pylint` to this list?

pllim: Added. Also skipped Appveyor this time since this PR does not touch Appveyor.

pllim: Ops... forgot the `[docs only]` in second commit. Do you want me to add it and force push?

bsipocz: yeah, please let's do that otherwise we get to our old age when it finishes (actually could also do a [skip ci] as I'm prepared to live dangerously and override the branch protection, as it already passed before).

pllim: Re-pushed!

pllim:
> That was on my sloppy review, sorry.

No worries! I had syntax highlighting on and I still missed it. My bad more than yours.

bsipocz:
Thanks @pllim!

PR #428, opened 2019-11-18
Remove usage of tee to avoid read-only issues on Windows

### LINKED ISSUE #427: Temporary  file is read only on windows
https://travis-ci.org/astropy/pytest-doctestplus/jobs/604227968#L221

./ci-helpers/travis/setup_dependencies_common.sh: line 93: /dev/fd/62: Read-only file system

It refers to this line: ``$revised_command > >(tee $_tmp_output_file) 2>&1`` from below

```
            python ci-helpers/travis/hack_version_numbers.py $_tmp_spec_conflicts "$@" > $_tmp_updated_conda_command
            revised_command=$(cat $_tmp_updated_conda_command)
            echo $revised_command
            # Try it; if it still has conflicts then just give up
            $revised_command > >(tee $_tmp_output_file) 2>&1
            _exitval="$?"
            if [[ -n $(grep "conflicts with explicit specs" $_tmp_output_file) ]]; then
                echo "STOPPING conda attempts because unable to resolve conda pinning issues"
                rm -f $_tmp_output_file
                return 1
            fi
        fi
```

cc @mwcraig @djhoese

djhoese:
Closes #427 

This follows the suggestions by @mwcraig to remove the usage of `tee` from `setup_dependencies_common.sh`. I still don't understand why Windows is making the file descriptor read-only for the pipe or what's going on and I'm still worried about this triggering CI time outs from no conda output, but if this makes "best case" Windows builds pass then so be it.

I'll be testing this with my own project and I'll report back later.

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -408,8 +408,8 @@ if [[ ! -z $CONDA_DEPENDENCIES ]]; then
     _tmp_output_file="tmp.txt"
     # do not exit on failure of the dry run because pip fallback may succeed
     set +e
-
-    conda install --dry-run $CONDA_DEPENDENCIES > >(tee $_tmp_output_file) 2>&1
+    conda install --dry-run $CONDA_DEPENDENCIES > $_tmp_output_file 2>&1
+    cat $_tmp_output_file
     set -e
     # 'grep' returns non-zero exit status if no lines match.
     if [[ ! -z $(grep "conflicts with explicit specs" $_tmp_output_file) ]]; then
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -90,8 +90,11 @@ function retry_on_known_error() {
             revised_command=$(cat $_tmp_updated_conda_command)
             echo $revised_command
             # Try it; if it still has conflicts then just give up
-            $revised_command > >(tee $_tmp_output_file) 2>&1
+            $revised_command > $_tmp_output_file 2>&1
             _exitval="$?"
+            # Keep the cat here...otherwise _exitval is always 0
+            # even if the conda install failed.
+            cat $_tmp_output_file
             if [[ -n $(grep "conflicts with explicit specs" $_tmp_output_file) ]]; then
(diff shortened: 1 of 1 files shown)


djhoese:
My stuff is still failing (at least setup isn't complaining about a read-only file system): https://travis-ci.org/ssec/sift/jobs/613285808

This looks like it is happening before my project-specific packages are being installed (I think).

bsipocz:
Hmm, weird. What happens if you switch to pip install everything?

bsipocz:
(also, that being the case, I've cancelled CI here for now. Happy to restart if you prefer)

djhoese:
I've made another branch `remove-tee2` on my fork where I just removed all of the retry stuff just to see what's really going on (raw conda command). I'm going to go offline in a bit and will test some more stuff tomorrow. I should see if I can get my hands on a Windows machine at work tomorrow and try some of these conda install commands just to see how it resolves the dependencies.

djhoese:
@bsipocz I'm still investigating. Any idea why travis on Windows would not be echoing the output of commands in the `setup_dependencies_common.sh` to the travis "console"? Even commands line `cat $PIN_FILE` aren't printing anything. I suppose the file could be empty, but even `conda install` commands don't show anything. It is output the `set -x` output but not much else.

https://travis-ci.org/ssec/sift/jobs/613295668

djhoese:
Ok so I did a lot of playing around on a Windows VM and found out that I'm running in to this issue: https://github.com/conda-forge/scipy-feedstock/issues/120

I'll try removing strict channel priority for my own project to verify that something better happens, but bottom line is that conda basically thinks the environment is unsolvable. @bsipocz I think you could start up the tests again.

You/we may want to document that hitting the 10m timeout on travis or other CIs is likely caused by an unsolvable environment that conda is trying its hardest to solve. If you need me to add that somewhere point me to where you think is best and I'll give it a shot.

djhoese:
My project passes. I had to make the Windows build flexible channel priority: https://travis-ci.org/ssec/sift/builds/613681043?utm_source=github_status&utm_medium=notification

bsipocz:
OK, so your project passes, this passes, I'm going ahead and merge it then.

Thank you @djhoese!

PR #429, opened 2019-12-14
Added script that can be used for tox builds

astrofrog:
This script installs and run tox, and optionally applies global patches if needed. The idea is to keep this super minimal even in the case where patches need to be applied.

I tried a slightly different approach of leaving it to the caller to actually run the tox command, and using the script to just modify the environment variables, but this didn't work because of issues with quotes in bash, which would require the final tox command to be convoluted.

For now, this script does not do any patching.

I've temporarily disabled all other Travis builds to make sure the one I added works, in case iteration is needed, without wasting CI time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
         - SETUP_CMD='test'
         - TEST_CMD='py.test test_env.py'
         - PYTHON_VERSION=3.6
+        - SETUP_CONDA=true
 
     matrix:
         - SETUP_CMD='egg_info' TEST_CMD='python --version'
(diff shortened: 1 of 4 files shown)


bsipocz:
Let me know whet this is ready for review.

astrofrog:
@bsipocz - ready for review, the Travis failure is unrelated

bsipocz:
Travis failure is with the helpers, so I'm rather inclined to patch rework the 4.0 core tag.

astrofrog:
@bsipocz - ok sounds good, that does indeed look like the issue that will be fixed in https://github.com/astropy/astropy-helpers/pull/506 - that PR should be ready to be merged

bsipocz:
OK, so having  the pre release wheels indeed solved the CI. 

I'll try to do a review later tonight.

bsipocz:
Thanks @astrofrog. And a note to myself that I need to stop saying things about timings that I end up apparently not keeping.

pllim:
`astropy` CI is still failing on `master`, so I just cleared the cache and restarting https://travis-ci.org/github/astropy/astropy/builds/662587719 . FYI

pllim:
Oh, right. `astropy` doesn't use `ci-helpers` anymore for Linux, right? 🤦‍♀

PR #430, opened 2019-12-31
Update the version numbers

bsipocz:
Neither astropy nor numpy made it to conda defaults yet, but if these are picked up for pip install without issue, I'll go ahead and merge 

(haven't yet updated the LTS version numbers)

### DIFF
--- a/test_env.py
+++ b/test_env.py
@@ -33,10 +33,10 @@ if 'TRAVIS_REPO_SLUG' in os.environ:
 
 
 if not LooseVersion(sys.version) < '3.5':
-    LATEST_ASTROPY_STABLE = '3.2.3'
+    LATEST_ASTROPY_STABLE = '4.0'
     # This is not for windows but appveyor only
     LATEST_ASTROPY_STABLE_WIN = '3.2.2'
-    LATEST_NUMPY_STABLE = '1.17'
+    LATEST_NUMPY_STABLE = '1.18'
 else:
     LATEST_ASTROPY_STABLE = '2.0.16'
     # This is not for windows but appveyor only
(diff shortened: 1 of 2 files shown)


djhoese:
This PR seems to be causing issues for OSX. I've tried it with python 3.6 and 3.7 right now and the "NUMPY_VERSION=stable" fails and says that numpy 1.18 is not available (even from conda-forge).

djhoese:
Forcing numpy to 1.17 fixes this for now.

bsipocz:
Indeed, the conda update is not yet merged on conda-forge: https://github.com/conda-forge/numpy-feedstock/pull/179

But even then, I would expect it to fall back on pip install, just as it does for the linux builds.

djhoese:
Yes, it was falling back for me but then some other issue was popping up. I only noticed the numpy version and went with forcing the version as the simplest solution. Let me see what the error was...

djhoese:
So the error is coming from cartopy which is being installed from pip as well (do all dependencies fallback to pip if one of them fails?):

https://travis-ci.org/ssec/polar2grid/jobs/632013555#L1156

Cartopy has a hard requirement on PROJ (the C library) being installed in the environment already and is failing because it doesn't find it. It gets installed as a conda dependency if the conda command would have succeeded.

bsipocz:
> (do all dependencies fallback to pip if one of them fails?)

no, they shouldn't, but there is a sub-optimal iteration to figure out which one is the culprit. It keeps the order of how the deps are listed, so I think listing proj as a dependency somewhere at the beginning, or moving pyproj in the dependency list before cartopy would solve the issue.

djhoese:
Alright, I'm not too worried about it for now. Numpy 1.17 is a good enough workaround for me right now. Thanks for explaining it.

adonath:
@bsipocz I'm seeing a similar issue in the Gammapy Travis: https://travis-ci.org/gammapy/gammapy/jobs/633830234#L1365

The conda install for `numpy==1.18` fails (as expected...) and it falls back to the pip install, but then fails with `ERROR: No matching distribution found for nomkl`. Do you maybe have any advice for a workaround? Otherwise I'd be also fine to stick with `numpy==1.17` for now....

bsipocz:
Yes, retrospect this might have been a rushed through update here. I don't understand why the nomlk issues doesn't come up for astropy. It definitely a bug, but I have very limited resources to hack on ci-helpers atm, so if you could stick with np==1.17 for a while, that would be great.

I've also pinged folks on the np 1.18 conda-forge PR to get it merged soonish...

adonath:
Thanks @bsipocz! I fixed numpy to 1.17 and will wait for the conda builds to be ready.

bsipocz:
@adonath - numpy 1.18.1 is now on the defaults channel, you can give it a go now, and ping me if any issues come up?

PR #435, opened 2020-01-08
Update LTS versions, and fix python version dependency

bsipocz:


djhoese:
I swear I'm plagued by dependencies and versions. This update seems to have broken things on appveyor. Not that it is ci-helpers fault but now that the default numpy is 1.17 instead of 1.16 I'm having weird interactions between numpy and libtiff. I don't expect you to do anything to fix it but wanted to mentioned it somewhere.

I'll see what I can figure out.

bsipocz:
Ouch. I'm sorry to hear that. Do you see the same issue with 1.18.1? it seems to be out on the defaults channel now.

Unfortunately setting the versions here will cause these chain reactions downstream, though if we don't set version in ci-helpers the same issues would come up probably anyway once/if those versions are picked up automatically by conda.

PR #437, opened 2020-01-23
Limit even more coverage 

bsipocz:
Maybe the third time will be the winning one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ matrix:
         # and we want the pip fallback to happen.
         - os: linux
           env: SETUP_CMD='test --coverage' PYTHON_VERSION=3.6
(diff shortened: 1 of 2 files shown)


djhoese:
I'm seeing some odd behavior in some of my builds that are having trouble finding dependencies because of these changes. Is there an issue explaining this? Why is this in ci-helpers and not in the astropy dependency list?

bsipocz:
Because it affects all of the astropy downstream packages that use the astropy test runner. That's tens if not hundreds of repos. This situation was exactly the motivation behind ci-helpers, but I'm happy have any workarounds that would work for everyone.

bsipocz:
e.g. hacky, but we could deal with having a `COVERAGE_VERSION` or similar

djhoese:
My understanding from the changes is that if I specify `pytest-cov coverage` in my dependencies it should get around this forced version, right?

You're right, this is astropy's project and thank you all for making it available. Is there an issue that I can see that describes why coverage>=5 breaks things in the test runner? I'm curious if I can help or if it is something I should avoid in my own projects.

PR #441, opened 2020-03-17
Allow python interpreter to be overriden and make sure pip and setuptools are recent

astrofrog:
Some further tweaks needed for astropy core.


bsipocz:
thanks @astrofrog!

PR #442, opened 2020-03-18
Use simple pinnings file for tox

astrofrog:
This changes the approach for tox to instead just have a single file in this repository that contains pinnings (in requirements.txt syntax) - see https://github.com/astropy/astropy/pull/10045 for how this is used.


bsipocz:
Call it maybe `pip_pinnings.txt`, or do we plan to use this with conda, too?

bsipocz:
Also, please add a sentence or two about it in the readme.

bsipocz:
Thanks!

PR #443, opened 2020-04-23
MNT: Remove [skip travis]

pllim:
Because `[skip travis]` is natively provided by Travis CI now, as @bsipocz pointed out at astropy/astropy#10186


bsipocz:
Thanks @pllim!

PR #444, opened 2020-04-23
Adding option for pytest dev version

bsipocz:


bsipocz:
I don't run CI again for the jobs that were untouched and passing before the last commit, but merge this instead.

PR #445, opened 2020-04-30
Remove pytest pinning

bsipocz:
pytest-doctestplus is released, so there is no more need to this version pinnnig


bsipocz:
I'm not exactly sure why the `'matplotlib>2.0,<=2.2.0'` requirement suddenly started to downgrade mpl to 2.0.2 (it first installs 2.2, but then downgrades it). At the end of the day, installing 2.0.2 is still within that limit range, so I'm not sure we should live with the assumption that for those cases we always get the most recent version. Anyhow, the issue appeared on master, too, since the last commit was made there.

pip_pinnings.txt
```diff
@@ -1 +1 @@
-pytest<5.4
+
```
pllim, line 1: I think this was that @astrofrog suggested on Slack?
```suggestion
pytest-doctestplus<0.6
```

pllim: Oh wait... We need to pin both now?

bsipocz: let's wait a bit more for a proper fix instead. pytest 5.4 is fine with doctestplus 0.6, but 0.5 is not working with 5.4, so we either need neither of them or both.

bsipocz:
failure is unrelated as noted above, and strictly speaking an issue with the test, so I'm merging this now.

pllim:
Re: Failure -- Not sure why but seems like it pulled down `matplotlib    2.0.2   np113py27_0` but the test is looking for 2.2.0.

bsipocz:
yes, but it used to pull the latest version from the given version range, but now sticks to the oldest. I suppose we better update the test matrix at some point, as the available conda packages might be limited for older pythons.

PR #446, opened 2020-04-30
Pin pytest-doctestplus

pllim:
To get rid of CI failures for astropy 4.1 feature freeze.


bsipocz:
Thanks @pllim. I'm going to close this though as the pytest-doctestplus release has been reverted, so no need for this (and either of the other two PR will provide a direct workaround in either astropy or doctestplus).

PR #449, opened 2020-05-05
TST: Fix build_sphinx job

pllim:
I don't question `conda` resolver's logic. Just trying to fix what I see in the log.


pllim:
Appveyor can be ignored, right? I think the `build_sphinx` job passed.

bsipocz:
OK. I don't understand why this is going on, but let's have it merged for the sake of a passing CI.

At some point a culling and upgrade of jobs is overdue in that test matrix.

Also, I'm not sure what happened with appveyor, it worked last week, but I have no desire to investigate right now but rather leave it a bit longer hoping it self resolves.

bsipocz:
thanks @pllim!

PR #453, opened 2020-07-28
Pin pytest<6

pllim:
Need to pin `pytest<6` until a new `pytest-doctestplus` with fix for astropy/pytest-doctestplus#118 is *released*. The patch is modeled after what @bsipocz did in #445 but in reverse (if that makes sense...).

cc @astrofrog


pllim:
Do we still support Appveyor? Should I care that Appveyor failed?

bsipocz:
If the appveyor failure is related, then yes it ideally should be fixed. If it's failing with something unrelated maybe the plug can be pulled in the near future (I'm reluctant to do it as quite a lot of projects used to use ci-helpers outside of astropy, so while this is a bit of a blast from the past at this point, as we see close to home, projects' leaderships may care little to nothing about DevOps leaving projects stuck with obsolete solutions).

pllim:
Re: Appveyor failures

conda` doesn't support Python 2.7 and 3.5 anymore, so those can probably be dropped from the matrix. The rest of the jobs are failing because `conda` is too old. So, I don't think they are related and I have no desire to fix them.

```
UnsatisfiableError: The following specifications were found
to be incompatible with the existing python installation in your environment:
Specifications:
  - conda=4.7 -> python[version='>=2.7,<2.8.0a0|>=3.7,<3.8.0a0|>=3.6,<3.7.0a0']
Your python: python=3.8
If python is on the left-most side of the chain, that's the version you've asked for.
```

bsipocz:
OK, let's see what a rebuild of the last merge says. 20 days days ago  it was still all green. If the rerun lights up the same as this PR, I'm pushing the merge button.

pllim:
Sounds like a plan. Thanks, @bsipocz !

bsipocz:
OK, so appveyor is very badly broken already. Something for another day....

pllim:
Thanks for checking and the merge!

Issue #456, opened 2020-09-11
Python 3.8 environments on travis hang


djhoese:
I just recently updated my travis to use python 3.8 and now the ci-helper's based conda install can't seem to get passed a certain installation step. From what I can tell it isn't related to the dependencies I'm installing.

Here is an example travis fail:

https://travis-ci.com/github/ssec/polar2grid/jobs/384028021

And the config:

<details>
```
language: generic
env:
  global:
    # Set defaults to avoid repeating in most cases
    - PYTHON_VERSION=$TRAVIS_PYTHON_VERSION
    - NUMPY_VERSION=stable
    - MAIN_CMD='python setup.py'
    - CONDA_DEPENDENCIES='xarray dask toolz Cython sphinx pillow matplotlib scipy pyyaml pyproj pyresample coveralls coverage codecov behave netcdf4 h5py h5netcdf gdal rasterio imageio pyhdf libtiff sphinx-argparse graphviz python-graphviz pycoast pytest zarr pylibtiff python-geotiepoints'
    - PIP_DEPENDENCIES='trollsift trollimage pyspectral pyorbital'
    - SETUP_XVFB=False
    - EVENT_TYPE='push pull_request'
    - SETUP_CMD='test'
    - CONDA_CHANNELS='conda-forge'
    - CONDA_CHANNEL_PRIORITY='strict'
matrix:
  include:
  - env: PYTHON_VERSION=3.8
    os: linux
  - env: PYTHON_VERSION=3.8
    os: osx
install:
  - git clone --depth 1 git://github.com/astropy/ci-helpers.git
  - source ci-helpers/travis/setup_conda.sh
  - pip install --no-deps git+https://github.com/pytroll/satpy.git
  - pip install -e . --no-deps
script:
  - python -m polar2grid.tests && cd doc && make html
```
</details>

Normally I would think this is an unsolvable conda environment but seeing as it doesn't seem to be including my dependencies I can't understand why.

astrofrog:
I wonder if we should have an option to use mamba, as it generally gets around these kind of issues and can help debugging - and it’s also just plain faster.

bsipocz:
+1 to try support mamba, conda can really be a pain sometimes.

PR #457, opened 2020-09-11
Add a way to opt in to using mamba for install commands

astrofrog:
This is an attempt to help address #456

### DIFF
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -170,9 +170,9 @@ is_eq_float="=[0-9]+\.[0-9]+"
 
 if [[ $MAMBA == True ]]; then
     conda install -c conda-forge mamba;
-    CONDA_INSTALL_COMMAND='mamba';
+    CONDA_INSTALL_COMMAND='mamba install';
 else
-    CONDA_INSTALL_COMMAND='conda';
+    CONDA_INSTALL_COMMAND='conda install';
 fi
 
 if [[ -z $PIP_FALLBACK ]]; then
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -168,6 +168,12 @@ is_number='[0-9]'
 is_eq_number='=[0-9]'
 is_eq_float="=[0-9]+\.[0-9]+"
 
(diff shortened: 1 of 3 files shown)


astrofrog:
Predictably, the install of mamba failed due to a conda error 😆

astrofrog:
@djhoese - could you try this out by using:

```
git clone --branch mamba https://github.com/astrofrog/ci-helpers
```

and let us know if that helps you diagnose the Python 3.8 issues?

astrofrog:
Ok so this *seems* to work - I had to bump the conda version when MAMBA=True which I think is ok since it won't change the default version for any existing users of ci-helpers who don't opt in to mamba.

djhoese:
Awesome! Thanks! Turns out the pylibtiff conda-forge package I was depending on hasn't been rebuilt for Python 3.8. Thanks!

https://github.com/conda-forge/pylibtiff-feedstock/issues/2

astrofrog:
Great, glad it worked!

astrofrog:
I need to document this - @bsipocz do you think MAMBA is ok as a environment variable name or would you prefer something else? Also I won’t bother to add this to Appveyor right?

travis/setup_dependencies_common.sh
```diff
@@ -190,7 +196,11 @@ fi
 # We pin the version for conda as it's not the most stable package from
 # release to release. Add note here if version is pinned due to a bug upstream.
 if [[ -z $CONDA_VERSION ]]; then
-    CONDA_VERSION=4.7.11
+    if [[ $MAMBA == True ]]; then
+        CONDA_VERSION=4.8.4
+    else
+        CONDA_VERSION=4.7.11
```
bsipocz, line 202: If the newer version works out of the box, it can be bumped (though if there are any issues, it's probably better not to go down that rabbit hole now)

bsipocz (approved):

bsipocz:
>  @bsipocz do you think MAMBA is ok as a environment variable name or would you prefer something else? Also I won’t bother to add this to Appveyor right?

- mamba sounds perfectly reasonable to me
- ignore appveyor, we agreed to ditch it

bsipocz:
Thanks @astrofrog!

Issue #458, opened 2020-09-13
Consider using micromamba


astrofrog:
Micromamba is the mamba equivalent of miniconda (see https://github.com/mamba-org/mamba#micromamba). We could imagine adding new ``setup_mamba*`` scripts which would set things up exclusively as opposed to using mamba just for installing as in #457. I imagine this could speed things up a bit. Because the scripts would be separate (except for the main one that sets up dependencies etc) this would be opt-in.

One limitation is that it looks like micromamba is not yet available for Windows (@wolfv, is that something that might exist in future?)

wolfv:
yes, I think we'll have a Windows build at the end of the week. I have just added `vcpkg` to conda-forge that we can use to bootstrap the static library dependencies of micromamba on Windows.

astrofrog:
Great, thanks!

PR #461, opened 2020-09-23
Fix mamba installation to work when installing dependencies

### LINKED ISSUE #460: Mamba fails in some environments
Still tracking this down but I enabled mamba on a vispy PR I'm working on and seeing:

```
Executing transaction: ...working... done

./ci-helpers/travis/setup_dependencies_common.sh: line 55: mamba: command not found

./ci-helpers/travis/setup_dependencies_common.sh: line 428: mamba: command not found

./ci-helpers/travis/setup_dependencies_common.sh: line 55: mamba: command not found

./ci-helpers/travis/setup_dependencies_common.sh: line 55: mamba: command not found
```

Here's the full job: https://travis-ci.com/github/vispy/vispy/jobs/386465377

I can tell from the log that mamba is definitely installed, but something is getting lost. Perhaps a weird mix of conda failing and ci-helpers trying to do pip as a fallback.

djhoese:
Close #460 

Mamba is not treated like the `conda` binary where it is copied to all sub-environments. This meant that once we activated our new `test` environment the `mamba` command was no longer available. We have to make sure `mamba` is installed as a dependency in this new environment if we want to use it.

CC @astrofrog who originally wrote #457

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,8 @@ environment variables
 * ``MAMBA``: if set to ``True``, conda packages
   will be installed with `mamba <https://github.com/mamba-org/mamba>`_, which is
   both faster than conda and gives more readable errors in cases where there are
-  conflicts.
+  conflicts. Note that if using ``CONDA_ENVIRONMENT`` you must include ``mamba``
+  as a dependency in addition to setting this to ``True``.
 
 * ``NUMPY_VERSION``: if set to ``dev`` or ``development``, the latest
   developer version of Numpy is installed along with Cython. If set to a
--- a/README.md
+++ b/README.md
@@ -76,8 +76,7 @@ environment variables
 * ``MAMBA``: if set to ``True``, conda packages
   will be installed with `mamba <https://github.com/mamba-org/mamba>`_, which is
   both faster than conda and gives more readable errors in cases where there are
-  conflicts. Note that if using ``CONDA_ENVIRONMENT`` you must include ``mamba``
-  as a dependency in addition to setting this to ``True``.
+  conflicts.
 
 * ``NUMPY_VERSION``: if set to ``dev`` or ``development``, the latest
   developer version of Numpy is installed along with Cython. If set to a
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -265,10 +265,10 @@ fi
 
 # install mamba and use it from now on
 if [[ $MAMBA == True ]]; then
-    CONDA_INSTALL_COMMAND='mamba install';
+    CONDA_INSTALL_COMMAND='mamba install'
     conda install -c conda-forge mamba
 else
-    CONDA_INSTALL_COMMAND='conda install';
+    CONDA_INSTALL_COMMAND='conda install'
 fi
 
 # CORE DEPENDENCIES
--- a/travis/setup_dependencies_common.sh
+++ b/travis/setup_dependencies_common.sh
@@ -169,10 +169,13 @@ is_eq_number='=[0-9]'
 is_eq_float="=[0-9]+\.[0-9]+"
 
 if [[ $MAMBA == True ]]; then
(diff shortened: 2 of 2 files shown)


bsipocz (approved):
all looks good, and should be good to go assuming travis status will be green

astrofrog:
Looks good to me too!

djhoese:
One of the test failures looks unrelated. The other though is an oversight on my part. We can't install `mamba` from defaults (yet) so we need to include the conda-forge channel. I see three options:

1. Move the mamba install to a separate step after the environment creation and use the conda-forge channel.
2. Only install mamba if conda-forge is in the channel list. Error or no-op otherwise.
3. Modify the mamba test to include the conda-forge channel and document in the README that using the mamba option requires the conda-forge channel to be included.

I personally am scared (from past experience) mixing defaults and conda-forge installations so I think I prefer 2 or 3.

bsipocz:
It may not be relevant here, but I have had very bad experiences using with conda-forge over default a couple of weeks ago, when a jupyterlab had to be set up with recent versions of e.g. tensorflow. Also, I recall previous issues with having conda-forge by default (which might have improved, but I still think many of the use cases prefers the curated nature of defaults). 

So, I would be inclined for opt 1), installing mamba only from conda-forge if it's not explicitly listed as a channel, but leave the rest as requested for the rest of the dependencies. I would be also fine with opt 2), but don't like opt 3) as it would mean the behaviour is different depending on using conda or mamba and I feel that is counter intuitive.

djhoese:
I think the majority of past issues with defaults and conda-forge (in general, not ci-helpers specifically) have always been mixing the two. I have had much better experiences with conda-forge now that I always specify a strict channel priority which is now what the conda-forge community recommends.

djhoese:
>  but don't like opt 3) as it would mean the behaviour is different depending on using conda or mamba and I feel that is counter intuitive.

This would still be up to the user though. Kind of like it is now where I've documented it as "if you'd like to use MAMBA and CONDA_ENVIRONMENT you need to include the mamba package". Now it would just mean "you also need to include conda-forge in your channel list". You know what, let me try option 1 and see what happens. It is already broken in master so what's the worst that can happen.

djhoese:
Oh and I just noticed the environment is created with `conda` regardless of mamba being specified. I think this could be made faster. Let's see what I can break...

bsipocz:
Ah, I'm sorry @djhoese, I misread option 3 and assumed it's a silent opt in to `conda-forge` when mamba is used. The proposed option, when read correctly sounds good to me.

djhoese:
I also had a misunderstanding of when/where things were happening. The original place in the script that mamba was installed was actually much earlier than it was used. In general it would probably be best if I (we?) could refactor ci-helpers to attempt all of the installations at once (environment creation time), but since this isn't the case installing mamba after the environment creation isn't that bad.  Oh...and it should remove the need to have mamba installed in your CONDA_ENVIRONMENT yaml file.

astrofrog:
This fix also looks good to me!

As a side note we could also imagine in MAMBA mode using micromamba instead of miniconda as described here: https://github.com/astropy/ci-helpers/issues/458 - but that might introduce additional complexity (what I would really like is if micromamba could be a drop in replacement for miniconda!)

djhoese:
So now that travis is passing I think this is "good enough". I'm not sure there is anything else that can be done without refactoring how the common setup works (all-in-one environment creation versus individual installs).

bsipocz:
Apparently I had an open tab with unsent comment and the follow-up of pushing the button from the day before yesterday. Sorry about that.

> OK, I suggest to go ahead with this, as is. I agree that a refactoring is a bit overdue at this point, this whole project grew rather organically and is now a bit of a monster hack. As I recall we have though about the refactoring a while ago, but deemed it low priority as we were dropping the usage in favour of e.g. tox.

bsipocz:
Thanks @djhoese!

PR #462, opened 2020-10-18
Add script to set up Python without conda

astrofrog:
Since the core package now uses choco to install Python on Windows and we might want to also avoid conda on MacOS X, I thought it would be handy to make a script that does all the set-up so we can use it across the organization as needed (the CI for the pytest plugins is having the same issues on Windows).

I am testing this out in https://github.com/astropy/astropy/pull/10871

### Why not use Homebrew on Mac?

It is quite slow and seems to download half the internet in some form of dependency hell.

### Why not use pyenv on Mac?

It downloads and builds Python from source so is very slow.

### Why use venv?

This is the easiest way to make sure we don't run into permissions issues and also makes available a ``python`` alias.

### Do we have to hard-code the bugfix mapping for MacOS X?

I could probably write a small script to try and do this automatically, but didn't want to spend too much time if we don't care.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ How to use
 
 ### Travis (with conda)
 
+*Note that you can also set up Python without conda using ci-helpers - see [here](#setting-up-python-without-conda-on-travis) for more details*
+
 Include the following lines at the start of the ``install`` section in
 ``.travis.yml``:
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ For the usage of the deprecated scripts see [Appveyor scripts README](https://gi
 How to use
 ----------
 
-### Travis
+### Travis (with conda)
 
 Include the following lines at the start of the ``install`` section in
 ``.travis.yml``:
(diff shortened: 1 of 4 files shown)


astrofrog:
@bsipocz - does this seem ok to you broadly speaking? If so will work on it some more and add docs etc to wrap up.

travis/setup_python.sh
```diff
@@ -0,0 +1,42 @@
...
+if [[ $TRAVIS_OS_NAME == osx ]]; then
+
+    if [[ $PYTHON_VERSION == 3.6 ]]; then
+        FULL_PYTHON_VERSION=3.6.9;
+    elif [[ $PYTHON_VERSION == 3.7 ]]; then
+        FULL_PYTHON_VERSION=3.7.9;
+    elif [[ $PYTHON_VERSION == 3.8 ]]; then
+        FULL_PYTHON_VERSION=3.8.6;
+    elif [[ $PYTHON_VERSION == 3.9 ]]; then
+        FULL_PYTHON_VERSION=3.9.0;
```
bsipocz, line 29: is the plan here to regularly revisit these numbers and update?

bsipocz (commented):
overall looks good

travis/setup_python.sh
astrofrog: I can write a small script to determine these automatically so we don't have to, but didn't want to spend time on it if you disagreed with it overall.

bsipocz: (It's OK if it's, but then maybe it's time to keep a tally of such updatable stuff (the current ones I know about are: latest versions of hard wired things in `LATEST...`, this python version, conda version)

bsipocz: I think it's reasonable enough to come back to it regularly, less sudden surprises that way (similar as with the conda verison)

astrofrog:
@bsipocz I added some docs including a list of variables to update regularly.

astrofrog:
https://github.com/astropy/astropy/pull/10871 seems to work, so ready for review

pllim:
I thought that we're slowly moving away from ci-helpers, given Linux jobs doesn't need ci-helpers anymore. Seems counter intuitive to add more stuff here just for Windows and OSX.

astrofrog:
To be clear we are moving away from conda ci helpers but we still want to use it for eg the pin file. Also I don't want to have to repeat this code in all Travis files :/

pllim (approved):
If this makes lives easier, why not.

Let's wait for @bsipocz 's approval before merge though.

pllim:
With this in, maybe we can start seriously think about getting rid of Appveyor.

astrofrog:
Yes I think since the appveyor script is broken anyway we could just display a message pointing people to using Windows on Travis instead.

README.md
```diff
@@ -18,7 +18,7 @@ a change is made.
 How to use
 ----------
 
-### Travis
+### Travis (with conda)
```
bsipocz, line 21: maybe link to the section without conda in the first sentence after this?

bsipocz (approved):

README.md
astrofrog: Sounds good will add

pllim: @astrofrog , after adding it, I think we can merge? Also if you really like green status, rebase and you shall have it.

astrofrog: Done! Because master is a protected branch we do need to wait for the CI to run again though.

PR #465, opened 2020-10-26
Install latest bugfix for Windows and fix venv creation and activation

astrofrog:
Follow-up of #462 - choco was installing the 3.x.0 versions instead of the latest bugfix and the venv creation and activation was broken.


astrofrog:
Looks like this is working fine based on Travis log so will go ahead and merge

Issue #469, opened 2021-03-09
MNT: Rename default branch from master to main


pllim:
Please rename your default branch from `master` to `main` as part of astropy/astropy-project#151 , preferably by 2021-03-22. Also a friendly reminder to check documentation, workflows, etc., and update them accordingly. Please don't forget to communicate this change to your users and stakeholders. To summarize:

- [x] Rename branch from `master` to `main`, preferably by 2021-03-22.
- [x] Update documentation, workflows, etc., accordingly. -- See #470
- [x] Communicate this change to your users and stakeholders. -- Slack `#infrastructure`

Once this is taken care of, you may close this issue.

*This is an automated issue. If this is opened in error, please let @pllim know!*

bsipocz:
Here I think we need to use the rename tool from `master` to `main`, but right after make a copy of the beach called `master`, so we don't break the all those projects that still use it using the link to the master branch.

(I don't support to keep a master copy for other repos, but here there won't be development on the main, default branch to diverge them anyway...)

pllim:
Yes, that makes sense. I kept `master` but the default is `main` now.

pllim:
Thanks for the review! 🔥

PR #470, opened 2021-03-09
MNT: Replace master with main

pllim:
This PR attempts to replace all mention of `master` to `main` in this repository. This should be reviewed and merged right after the maintainer has renamed the default branch.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

This is part of #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ service are no longer supported. Please use the
 [Windows build on Travis](https://docs.travis-ci.com/user/reference/windows/)
 instead.
 
-For the usage of the deprecated scripts see [Appveyor scripts README](https://github.com/astropy/ci-helpers/blob/master/appveyor/README.md).
+For the usage of the deprecated scripts see [Appveyor scripts README](https://github.com/astropy/ci-helpers/blob/main/appveyor/README.md).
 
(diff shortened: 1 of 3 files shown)


pllim:
Does depend on APE repo renaming too.

bsipocz:
I would say this is a good place to start to break the almost circular dependencies, so 👍 on merging now.

pllim:
Hmm... If the renaming also does auto-redirect, then maybe we should have just renamed it instead of making a copy. But I guess that ship has sailed for this repo.

https://github.com/astropy/astropy-APEs/issues/66#issuecomment-795833630

bsipocz:
Yep, same mistake I made for astroml, but I suppose it's better to be safe than sorry (and we can still delete the copy once you get an answer for the helpdesk ticket)

pllim:
From https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/ , it says, "This change only affects view links; other types of links (like edit links and blame links) don't redirect."

Do you think we can just go ahead and do #471 now and see what happens?

